PR #22, opened 2020-02-14
Use JSON validation from yaup

rvansa:
@willr3 I have removed the jitpack dependency; could you release yaup 0.5 and **then** integrate this?


johnaohara:
I pushed a 0.5 release of yaup, so this is ready to integrate

rvansa:
Self-merging.

PR #79, opened 2021-09-30
Include Java client in main repo

johnaohara:
This PR fixes an API interface definition and pulls the java client into the main repo.

I am not currently happy with the way the test suite bootstraps, it is currently pulling a Horreum image from quay.io to test the client against. Tests should be run against the newly built horreum-backend


horreum-java-client/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
```diff
@@ -0,0 +1,161 @@
...
+                    .withLocalCompose(true)
+                    .withExposedService("postgres_1", 5432)
+            ;
+            environment.start();
+            //wait for Horreum infrastructure to spin up
+            log.info("Waiting for Horreum infrastructure to start");
+            Optional<ContainerState> optionalContainer = environment.getContainerByServiceName("horreum_1");
+            if (optionalContainer.isPresent()) {
+                ContainerState horreumState = optionalContainer.get();
+                while (!horreumState.getLogs(OutputFrame.OutputType.STDOUT).contains("started in")) {
```
rvansa, line 93: You could make this more responsive with
```
.withLogConsumer(new ToStringConsumer() {
   void accept(OutputFrame frame) {
      super.accept(frame);
      if (toUtf8String().contains("started in")) {
         latch.countDown();
      }
   }
})
```

horreum-java-client/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
```diff
@@ -0,0 +1,161 @@
...
+    @BeforeAll
+    public static void beforeClass() throws Exception {
+        initialiseRestClients();
+
+        log.info("Starting Infra: " + START_HORREUM_INFRA);
+        if (START_HORREUM_INFRA) {
+
+            environment = new DockerComposeContainer(new File("src/test/resources/testcontainers/docker-compose.yml"))
+                    .withLocalCompose(true)
+                    .withExposedService("postgres_1", 5432)
```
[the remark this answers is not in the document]
rvansa: Also the names of services/containers could be prefixed by `test-`.

horreum-java-client/src/test/resources/env.properties
```diff
@@ -0,0 +1,30 @@
...
+#horreum.keycloak.base-url=http://127.0.0.1:8180 #podman
+horreum.keycloak.base-url=http://172.17.0.1:8180
+
+horreum.username=user
+horreum.password=secret
+horreum.client_secret=96af3ebe-bf3c-4c70-8ca9-ee952ab42fec
+
+horreum.base-url=http://127.0.0.1:8082
+horreum.client-id=horreum-ui
+keycloak.realm=horreum
```
rvansa, line 14: horreum.keycloak.realm

horreum-java-client/src/test/resources/testcontainers/horreum.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/sh
```
rvansa, line 1: I really dislike duplicating the 'main' scripts. Could we use a symlink here?

rvansa (commented):
Thanks for the migration. If you don't have time for any of the suggestions we could get this in and handle that later on.

Btw. do we need to call the module horreum-**java**-client? Maybe horreum-client would be sufficient? (if we create e.g. Python client this will definitely live in its own repo).

horreum-java-client/pom.xml
```diff
@@ -0,0 +1,188 @@
...
+        <resteasy-client.version>3.13.2.Final</resteasy-client.version>
+        <keycloack-admin-client.version>14.0.0</keycloack-admin-client.version>
+    </properties>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>io.hyperfoil.tools</groupId>
+            <artifactId>horreum-api</artifactId>
+            <version>0.1-SNAPSHOT</version>
```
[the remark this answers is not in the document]
johnaohara: will update

horreum-java-client/pom.xml
```diff
@@ -0,0 +1,188 @@
...
+                </exclusion>
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-core</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>io.hyperfoil.tools</groupId>
```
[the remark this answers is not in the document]
johnaohara: This could be removed if I think about the deps more.  In the parent pom we exclude the graal sdk, and therefore the graalsdk is not pulled in as a dependency; however the client will not compile without the graal sdk on the classpath. 

I am thinking that we probably should not expose the Json class from yaup in the horreum api, due to the number of dependencies that it pulls into the project.  Maybe we should look at replacing Json in the api with a more lightweight class and using yaup Json internally if we need to?  

Also, I need to create a custom serializer for the yaup Json class, that we could possibly remove if we choose a different class to represent the JSON payload

horreum-java-client/src/main/java/io/hyperfoil/tools/HorreumClient.java
```diff
@@ -0,0 +1,140 @@
...
+            this.horreumPassword = horreumPassword;
+            return this;
+        }
+
+        public Builder clientId(String clientId) {
+            this.clientId = clientId;
+            return this;
+        }
+
+        public Builder clientSecret(String clientSecret) {
```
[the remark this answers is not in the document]
johnaohara: What are the chances that the config in keycloack could change? Would it be wise to make clientSecret optional?

horreum-java-client/src/main/java/io/hyperfoil/tools/HorreumClient.java
```diff
@@ -0,0 +1,140 @@
...
+import io.hyperfoil.tools.yaup.json.Json;
+import org.jboss.resteasy.client.jaxrs.ResteasyClient;
+import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
+import org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl;
+
+import javax.ws.rs.core.UriBuilder;
+
+public class HorreumClient {
+
+    public AlertingService alertingService;
```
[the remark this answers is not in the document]
johnaohara: will do

horreum-java-client/src/test/java/io/hyperfoil/tools/HorreumClientTest.java
```diff
@@ -0,0 +1,77 @@
...
+        assertNotNull(test);
+        assertEquals(getProperty("horreum.test.description"), test.description);
+
+        RunService.RunCounts runCounts = horreumClient.runService.runCount(test.id);
+        assertEquals(0, runCounts.total);
+
+        try {
+            horreumClient.runService.addRunFromData("$.start", "$.stop", test.name, getProperty("horreum.test.owner"), Access.PUBLIC, null, null, "test", payload);
+        } catch (BadRequestException badRequestException) {
+            fail(badRequestException.getMessage() + (badRequestException.getCause() != null ? " : " + badRequestException.getCause().getMessage() : ""));
```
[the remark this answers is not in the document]
johnaohara: this was here to allow me to see what the cause of the bad request was. for example: https://github.com/Hyperfoil/Horreum/blob/master/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java#L319

If there is an exception thrown in the server (e.g. "Cannot identify test name.", "Failed to find test "  etc) this is encoded in the payload of the response.  If I leave the test framework to handle the exception, it does not print out that payload and I have no clue as to **why** it was a bad request. This may be due to the way ServiceException is building the response (https://github.com/Hyperfoil/Horreum/blob/fbccd421931d65aa19a9131103e83458b01e6d88/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ServiceException.java#L26), but I did not look at validating that behaviour as part of this PR

horreum-java-client/src/test/java/io/hyperfoil/tools/HorreumClientTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+        String jsonContemt;
+        Json payload = null;
+
+        try (InputStream inputStream = Thread.currentThread().getContextClassLoader().getResourceAsStream("data/config-quickstart.jvm.json");) {
+            jsonContemt = new BufferedReader(new InputStreamReader(inputStream))
+                    .lines().collect(Collectors.joining(" "));
+            payload = Json.fromString(jsonContemt);
+        } catch (IOException ioException) {
+            ioException.printStackTrace();
```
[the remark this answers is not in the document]
johnaohara: missed that, thanks

horreum-java-client/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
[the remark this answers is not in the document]
johnaohara: I can update the port numbers to use ports that will not conflict

iirc, testcontainers prefixes the container name with "testcontainer-{UID}-postrges_1" or something similar to avoid naming conflicts, but will dble check this behaviour

horreum-java-client/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
johnaohara: yes, this is suboptimal :)

horreum-java-client/src/test/resources/testcontainers/horreum.sh
johnaohara: i will clean this up

horreum-java-client/pom.xml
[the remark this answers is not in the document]
rvansa: https://github.com/Hyperfoil/Horreum/issues/80

I would not block this PR though unless using YAUP requires further development. Let's integrate it with the 'garbage' dependencies and then get rid of YAUP in one shot.

horreum-java-client/src/main/java/io/hyperfoil/tools/HorreumClient.java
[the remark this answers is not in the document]
rvansa: We are in control. Keycloak offers 'Confidential', 'Public' and 'Bearer only' access types; the confidential flow being default - that's why I've chosen it first.

> Public clients are perfectly fine so long as HTTPS is strictly enforced and you are very strict about what redirect URIs are registered for the client.

https://www.keycloak.org/docs/latest/server_admin/index.html#_oidc-auth-flows

So I guess we could stick to public-only, but require the URL to use TLS unless explicitly disabled?

horreum-client/src/test/java/io/hyperfoil/tools/HorreumClientTest.java
```diff
@@ -0,0 +1,77 @@
...
+            horreumClient.runService.addRunFromData("$.start", "$.stop", test.name, getProperty("horreum.test.owner"), Access.PUBLIC, null, null, "test", payload);
+        } catch (BadRequestException badRequestException) {
+            fail(badRequestException.getMessage() + (badRequestException.getCause() != null ? " : " + badRequestException.getCause().getMessage() : ""));
+        }
+
+    }
+
+    @Test
+    @Order(2)
+    public void ConfigQuickstartTest() {
```
rvansa, line 55: I wonder what difference is between these tests? Both test the `addRunFromData` endpoint. Whether you create the JSON payload manually or read it from file should be irrelevant.

horreum-client/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
```diff
@@ -0,0 +1,219 @@
...
+    private static final String HORREUM_CLIENT_SECRET;
+    private static final Integer HORREUM_TEST_PORT_OFFSET;
+
+    private static final String CONTAINER_HOST_IP;
+
+    private static final Integer HORREUM_HTTP_PORT = 8082;
+    private static final Integer HORREUM_HTTPS_PORT = 8443;
+    private static final Integer GRAFANA_HTTP_PORT = 4040;
+    private static final Integer POSTGRES_PORT = 5432;
+    private static final Integer KEYCLOACK_PORT = 8180;
```
rvansa, line 46: You have a typo at multiple places: `KEYCLOACK` -> `KEYCLOAK`.

horreum-client/src/test/resources/testcontainers/horreum.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
```
rvansa, line 1: We've discussed that it would be better not to duplicate these scripts.

johnaohara: I have not finished cleaning this up yet. They are not strictly duplicated so am working on unifying their behaviour atm

horreum-client/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
johnaohara: Thats what happens when you copy/paste an incorrect spelling. Will fix

horreum-client/src/test/resources/testcontainers/horreum.sh
rvansa: Sorry, no rush. I just thought it's up for the review when you've pushed the changes.
Btw. when you're done please squash the changes and rebase.

horreum-java-client/src/test/resources/testcontainers/horreum.sh
johnaohara: I could not get symlinks to work inside a container, so have used maven-resources-plugin to copy required resources into the target dir for the client tests

PR #97, opened 2021-12-09
Fix script permissions when client testsuite is used as a dependancy

### LINKED ISSUE #96: client testsuite scripts do not have correct mode set when referenced from dependent project
When using the client testsuite from another project, the docker-compose command fails with the following error when users try to bootstrap the testsuite docker environment;

```
INFO: ERROR: for r1zertyz9cas_keycloak_1  Cannot start service keycloak: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "/cwd/keycloak.sh": permission denied: unknown
INFO: ERROR: for keycloak  Cannot start service keycloak: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "/cwd/keycloak.sh": permission denied: unknown
```
The shell scripts are extracted from the horreum-client test jar, but they do not have the correct permissions;

```
$ ls -la target/docker-compose/
total 120
drwxrwxr-x.  3 johara johara  4096 Dec  9 19:31 .
drwxrwxr-x. 14 johara johara  4096 Dec  9 19:21 ..
-rw-r--r--.  1 johara johara  5645 Dec  9 19:31 app-init.sh
-rw-r--r--.  1 johara johara  1895 Dec  9 19:31 disable-https.sh
-rw-r--r--.  1 johara johara  3228 Dec  9 19:31 docker-compose.yml
-rw-r--r--.  1 johara johara  1991 Dec  9 19:31 enable-https.sh
-rw-r--r--.  1 johara johara   163 Dec  9 19:31 grafana.sh
drwxrwxr-x.  2 johara johara  4096 Dec  9 19:25 horreum-backend
-rw-r--r--.  1 johara johara  1118 Dec  9 19:31 horreum-compose.yml
-rw-r--r--.  1 johara johara 71816 Dec  9 19:31 keycloak-horreum.json
-rw-r--r--.  1 johara johara   612 Dec  9 19:31 keycloak.sh
-rw-r--r--.  1 johara johara  2942 Dec  9 19:31 podman-compose.yml

```

johnaohara:
fixes #96

This implementation is not very elegant, but operations to check file permissions on a ZipFile throw an UnsupportedOperationException, so there is no easy way to discover the original file permissions.

The other option is to remove the file copy code in `prepareDockerCompose()` and use the maven-dependency-plugin to unpack dependencies in the project importing the testsuite. This would retain the correct permissions, but require each project to declare the maven-dependency-plugin in it's pom.xml

### DIFF
--- a/horreum-client/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
+++ b/horreum-client/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
@@ -188,8 +188,12 @@ public class HorreumTestBase {
         Files.walkFileTree(source, Collections.emptySet(), 1, new SimpleFileVisitor<>() {
             @Override
             public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
-                Files.copy(file, Path.of("target/docker-compose/", file.getFileName().toString()),
+                Path destPath = Path.of("target/docker-compose/", file.getFileName().toString());
+                Files.copy(file, destPath,
                       StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.COPY_ATTRIBUTES);
+                if(file.toString().endsWith(".sh")){
+                    destPath.toFile().setExecutable(true);
+                }
                 return FileVisitResult.CONTINUE;
             }
         });


rvansa:
I have no problem with the proposed solution, especially since it's only in the testsuite. Quick google told me that others have problems with ZipFileSystem and permissions, too: https://stackoverflow.com/questions/35838221/java-zipfilesystem-attributes

Issue #124, opened 2022-02-09
Solve never ending build task in Eclipse.


whitingjr:
Eclipse 4.21 gets into an infinitute build cycle. The maven-antrun-plugin causes this despite the default goal is run.
 This makes the IDE unusable.

PR #125, opened 2022-02-09
Solve issue with Eclipse build triggering repeatedly.

whitingjr:
This PR changes the configuration for the life-cycle mapping. To explicitly define the run goal as the default.


rvansa:
Thanks, does this solve #124 ? If it's so please mark the issue as resolved.

PR #126, opened 2022-02-09
Enable email missing run notifications to be muted. 

whitingjr:
Reuse the SubscrptionServiceImpl.get method to reduce code duplication. Added testcase for Watch update.

 This PR adds a checkbox to the UI to disable sending notification emails to users.

### DIFF
--- a/example-data.sql
+++ b/example-data.sql
@@ -46,7 +46,7 @@ INSERT INTO datapoint (id,variable_id,runid,timestamp,value) VALUES (15,13,2,'20
 INSERT INTO datapoint (id,variable_id,runid,timestamp,value) VALUES (16,13,3,'2020-07-30 12:00:54.678',205);
 INSERT INTO change(id, confirmed, description, variable_id, runid, timestamp) VALUES (17, false, 'Test annotation <a href="http://localhost:3000/" target="_parent">with a link</a>', 13, 3, '2020-07-30 12:00:54.678');
 
-INSERT INTO watch(id, testid, mutemissingruns) VALUES (18, 2, false);
+INSERT INTO watch(id, testid, missingruns) VALUES (18, 1, false);
 INSERT INTO watch_users(watch_id, users) VALUES (18, 'user');
 INSERT INTO watch_teams(watch_id, teams) VALUES (18, 'dev-team');
 INSERT INTO notificationsettings(id, isteam, name, method, data) VALUES (19, false, 'user', 'email', 'john@example.com');
--- a/example-data.sql
+++ b/example-data.sql
@@ -46,7 +46,7 @@ INSERT INTO datapoint (id,variable_id,runid,timestamp,value) VALUES (15,13,2,'20
 INSERT INTO datapoint (id,variable_id,runid,timestamp,value) VALUES (16,13,3,'2020-07-30 12:00:54.678',205);
 INSERT INTO change(id, confirmed, description, variable_id, runid, timestamp) VALUES (17, false, 'Test annotation <a href="http://localhost:3000/" target="_parent">with a link</a>', 13, 3, '2020-07-30 12:00:54.678');
 
-INSERT INTO watch(id, testid) VALUES (18, 2);
+INSERT INTO watch(id, testid, mutemissingruns) VALUES (18, 2, false);
 INSERT INTO watch_users(watch_id, users) VALUES (18, 'user');
 INSERT INTO watch_teams(watch_id, teams) VALUES (18, 'dev-team');
 INSERT INTO notificationsettings(id, isteam, name, method, data) VALUES (19, false, 'user', 'email', 'john@example.com');
--- a/horreum-backend/src/main/resources/db/changeLog.xml
+++ b/horreum-backend/src/main/resources/db/changeLog.xml
@@ -538,6 +538,9 @@
             <column name="testid" type="integer">
                 <constraints nullable="false"/>
             </column>
+            <column name="mutemissingruns" type="boolean">
+                <constraints nullable="false"/>
+            </column>
         </createTable>
         <addUniqueConstraint tableName="watch" columnNames="testid"/>
         <createTable tableName="watch_users">
@@ -645,6 +648,7 @@
             CREATE POLICY watch_delete ON watch FOR DELETE USING (has_role('horreum.alerting'));
             CREATE POLICY watch_teams_policies ON watch_teams FOR ALL USING (has_role('horreum.alerting') OR has_role(teams));
             CREATE POLICY watch_users_policies ON watch_users FOR ALL USING (has_role('horreum.alerting') OR has_role(users));
+            ALTER POLICY watch_update ON watch USING (true);
         </sql>
     </changeSet>
 
--- a/horreum-backend/src/main/resources/db/changeLog.xml
+++ b/horreum-backend/src/main/resources/db/changeLog.xml
@@ -538,9 +538,6 @@
             <column name="testid" type="integer">
                 <constraints nullable="false"/>
             </column>
-            <column name="mutemissingruns" type="boolean">
-                <constraints nullable="false"/>
-            </column>
         </createTable>
         <addUniqueConstraint tableName="watch" columnNames="testid"/>
         <createTable tableName="watch_users">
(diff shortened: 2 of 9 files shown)


horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TestServiceTest.java
```diff
@@ -57,4 +63,58 @@ public void testCreateDelete(TestInfo info) {
...
+      Test test = createExampleTest(getTestName(info));
+      Test response = createTest(test);
+      Watch watch = new Watch();
+      test = checkTestAsExpected(test, response, watch);
+      watch = checkWatchAsExpected(test, watch);
+      watch.mutemissingruns = true;
+      updateWatch(test, watch);
+   }
+
+   private Test checkTestAsExpected(Test test, Test response, Watch watch) {
```
rvansa, line 79: The method name suggest validation but you're actually adding a watch...

webapp/src/domain/tests/Subscriptions.tsx
```diff
@@ -47,6 +49,10 @@ export default function Subscriptions(props: SubscriptionsProps) {
     const [watchingTeams, setWatchingTeams] = useState<ReactElement[]>([])
 
     const [reloadCounter, setReloadCounter] = useState(0)
+    const [isCheckedMissingRuns, setIsCheckedMissingRuns] = useState(props.mutemissingruns == "false")
+    const handleChangeMissingRuns = (e: React.ChangeEvent<HTMLInputElement>) => {
+        setIsCheckedMissingRuns(e.target.value == "false")
```
rvansa, line 54: Use `===` for string comparison.

webapp/src/domain/tests/Subscriptions.tsx
```diff
@@ -8,11 +8,13 @@ import { info, teams, User } from "../user/api"
 import { alertAction } from "../../alerts"
 import { teamToName, useTester } from "../../auth"
 import { TabFunctionsRef } from "../../components/SavedTabs"
+import { Checkbox } from "@patternfly/react-core"
 import UserSearch from "../../components/UserSearch"
 
 type SubscriptionsProps = {
     testId: number
     testOwner?: string
+    mutemissingruns: string
```
rvansa, line 17: why is this a string and not a `boolean`?

webapp/src/domain/tests/Subscriptions.tsx
```diff
@@ -139,6 +147,19 @@ export default function Subscriptions(props: SubscriptionsProps) {
                     props.onModified(true)
                 }}
             />
+            <br />
+            <Divider />
+            <br />
+            <Checkbox
+                onChange={(checked, ev) => {
+                    setIsCheckedMissingRuns(checked)
+                    ev.currentTarget.value = checked.toString()
```
rvansa, line 156: Are you sure this line is necessary?

Btw., there's https://www.patternfly.org/v4/components/switch component (the choice is up to you, but I find Switch nicer when it's one option alone). Check out https://www.patternfly.org/v4/guidelines/usage-and-behavior#selecting-between-options-on-a-form

rvansa (changes requested):

horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -538,6 +538,9 @@
             <column name="testid" type="integer">
                 <constraints nullable="false"/>
             </column>
+            <column name="mutemissingruns" type="boolean">
```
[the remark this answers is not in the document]
whitingjr: ok I'll add a change set instead.
correct, I blew away the prior database.

horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -645,6 +648,7 @@
             CREATE POLICY watch_delete ON watch FOR DELETE USING (has_role('horreum.alerting'));
             CREATE POLICY watch_teams_policies ON watch_teams FOR ALL USING (has_role('horreum.alerting') OR has_role(teams));
             CREATE POLICY watch_users_policies ON watch_users FOR ALL USING (has_role('horreum.alerting') OR has_role(users));
+            ALTER POLICY watch_update ON watch USING (true);
```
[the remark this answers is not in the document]
whitingjr: It wasn't obvious (being new to this project) in this implementation but I'll take variable_update policy as an example on what's permitted.

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TestServiceTest.java
whitingjr: Yes renamed to be more detailed.

webapp/src/domain/tests/Subscriptions.tsx
whitingjr: ok

webapp/src/domain/tests/Subscriptions.tsx
whitingjr: ok I'll try and get this switched over.

webapp/src/domain/tests/Subscriptions.tsx
whitingjr: changed

webapp/src/domain/tests/Subscriptions.tsx
whitingjr: yes I found the code for onChange handler necessary. Removing it causes nothing to happen when the UI object is clicked.

example-data.sql
```diff
@@ -46,7 +46,7 @@ INSERT INTO datapoint (id,variable_id,runid,timestamp,value) VALUES (15,13,2,'20
 INSERT INTO datapoint (id,variable_id,runid,timestamp,value) VALUES (16,13,3,'2020-07-30 12:00:54.678',205);
 INSERT INTO change(id, confirmed, description, variable_id, runid, timestamp) VALUES (17, false, 'Test annotation <a href="http://localhost:3000/" target="_parent">with a link</a>', 13, 3, '2020-07-30 12:00:54.678');
 
-INSERT INTO watch(id, testid) VALUES (18, 2);
+INSERT INTO watch(id, testid, mutemissingruns) VALUES (18, 2, false);
```
rvansa, line 49: Shouldn't this be just `missingruns`?

TBH it seems to me that `missingruns = true` is not very clear name for the fact that the missing runs notifications are to be **muted** (I would expect the opposite meaning). Maybe at least add a comment to the entity if you prefer it this way.

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TestServiceTest.java
rvansa: I still see the old name, `checkTestAsExpected`...

webapp/src/domain/tests/Subscriptions.tsx
rvansa: Fair enough, I can try to find out what's happening when this gets integrated.

example-data.sql
whitingjr: Yes it should be. Thanks for spotting that.
 Changing the field name made sense when it came to the UI Switch label. Writing a label that was obvious sense was not easy. A double negative. ie, "turn it on" to cease notifications.
 I've changed it to default to true. So that users can see they need to "turn it off" to cease notifications.

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TestServiceTest.java
whitingjr: Thought I had updated that.

webapp/src/domain/tests/Subscriptions.tsx
```diff
@@ -47,6 +50,7 @@ export default function Subscriptions(props: SubscriptionsProps) {
     const [watchingTeams, setWatchingTeams] = useState<ReactElement[]>([])
 
     const [reloadCounter, setReloadCounter] = useState(0)
+    const [isCheckedMissingRuns, setIsCheckedMissingRuns] = useState(props.missingruns == true)
```
rvansa, line 53: In Javscript/Typescript you should always use `===` for equality comparison. Though, here I think you don't need the property at all since you always set it to `false` - you could remove it alltogether and just use `useState(false)`.

rvansa (commented):
I can't see if 'changing the default' was applied - the default database value is `false` and if I am not mistaken the UI will show the switch off by default.

rvansa:
Also, it would be great if the name of variables conveyed the function even on the backend side (I would expect `missingruns` to be a collection, without the comments I would have little idea what `true` really means without looking where this is applied).

whitingjr:
**missingrunnotifications** being suitable?

rvansa:
Yes, preferably camel-cased. Assuming that `true` (default) means sending notifications.

Btw., I am asking a bit late, but what is exactly the difference between having this switched off and having empty `stalenessSettings`? Do you want to disable this only temporarily for all combinations of tags?

PR #130, opened 2022-02-11
Avoid writing to response that's aleady committed.

### LINKED ISSUE #129: Avoid writing to already committed response.
The RouteFilter attempts to resurrect a meaningful response for a 404.
 This causes a distracting error that the response is already committed.

whitingjr:
closes #129


rvansa:
Great catch, thanks!

PR #141, opened 2022-02-21
Add group to missing variable notification

dupliaka:
Show in notification groups to identify results

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java
@@ -364,7 +364,7 @@ public class AlertingServiceImpl implements AlertingService {
                if (recalculation != null) {
                   recalculation.runsWithoutValue.add(run.id);
                }
-               missingValueVariables.add(var.name);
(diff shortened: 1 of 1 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java
```diff
@@ -423,6 +423,10 @@ private void emitDatapoints(Run run, boolean notify, boolean debug, Recalculatio
       }
    }
 
+   private String getMissingVariableString(String name, String group) {
+      return (group.isEmpty()) ? name : group + "/" + name;
```
rvansa, line 427: This does not handle the `group == null` case.

rvansa (changes requested):
This could work, bug given that `dataPoint.variable` is initialized via em.getReference(...)` reading `group` property would result in an unnecessary DB lookup. What about querying the group name already in `LOOKUP_VARS` and adding that to `VarInfo`? That way you could have `getMissingVariableString` as method straight on `VarInfo`.

Sorry if I am nitpicking :)

rvansa:
Hi @dupliaka are you going to address the comments?

rvansa (approved):

PR #142, opened 2022-02-23
Added DataSet entity and configuration to changeLog.

### LINKED ISSUE #137: Create `DataSet` entity
a mirror of current Run, plus linking (1:N) to the original Run

whitingjr:
Fixes #137

### DIFF
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/DataSet.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/DataSet.java
@@ -1,0 +1,67 @@
+package io.hyperfoil.tools.horreum.entity.json;
+
+import java.time.Instant;
+
+import javax.persistence.CascadeType;
+import javax.persistence.Column;
+import javax.persistence.ConstraintMode;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.ForeignKey;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.SequenceGenerator;
+
+import org.hibernate.annotations.Type;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.JsonNode;
+
+import io.quarkus.runtime.annotations.RegisterForReflection;
+import io.smallrye.common.constraint.NotNull;
+
+@Entity(name="dataset")
+@RegisterForReflection
+/**
+ * Purpose of this object is to represent derived run data.
+ */
+public class DataSet extends OwnedEntityBase {
+
+   public static final String FIND_RUN_BY_ID = "DataSet.findRunById";
+   public static final String EVENT_NEW = "dataset/new";
+   public static final String EVENT_TRASHED = "dataset/trashed";
+
+   @Id
+   @SequenceGenerator(
+         name="datasetSequence",
+         sequenceName="dataset_id_seq",
+         allocationSize=1)
+   @GeneratedValue(strategy=GenerationType.SEQUENCE,
+         generator= "datasetSequence")
+   public Integer id;
+
+   @NotNull
+   @Column(name="start", columnDefinition = "timestamp")
+   public Instant start;
+
+   @NotNull
+   @Column(name="stop", columnDefinition = "timestamp")
+   public Instant stop;
+
+   public String description;
+
+   @NotNull
+   public Integer testid;
+
+   @NotNull
+   @Type(type = "io.hyperfoil.tools.horreum.entity.converter.JsonUserType")
+   public JsonNode data;
+
+   @ManyToOne(cascade = CascadeType.DETACH, fetch = FetchType.LAZY)
+   @JoinColumn(name = "runid")
+   @JsonIgnore
+   public Run run;
+}
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/Run.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/Run.java
@@ -5,14 +5,17 @@ import io.quarkus.runtime.annotations.RegisterForReflection;
 import org.hibernate.annotations.DynamicUpdate;
 import org.hibernate.annotations.Type;
 
+import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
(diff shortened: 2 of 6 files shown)


horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/DataSet.java
```diff
@@ -0,0 +1,65 @@
...
+   @NotNull
+   public Integer testid;
+
+   @NotNull
+   @Type(type = "io.hyperfoil.tools.horreum.entity.converter.JsonUserType")
+   public JsonNode data;
+
+   @NotNull
+   @Column(columnDefinition = "boolean default false")
+   public boolean trashed;
```
rvansa, line 61: We don't need trashed datasets - a run that is trashed shouldn't map to any datasets, and when it's untrashed the datasets should be created again.

horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -2158,5 +2158,42 @@
...
+            </column>
+            <column name="access" type="integer">
+                <constraints nullable="false" />
+            </column>
+            <column name="token" type="text" />
+            <column name="trashed" type="boolean" defaultValue="false" />
+        </createTable>
+        <sql>
+            GRANT ALL ON SEQUENCE dataset_id_seq TO "${quarkus.datasource.username}";
+            GRANT SELECT, INSERT, DELETE, UPDATE ON TABLE dataset TO "${quarkus.datasource.username}";
```
rvansa, line 2195: You should also `ALTER TABLE dataset ENABLE ROW LEVEL SECURITY;` and create policies checking access to the table.

horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -2158,5 +2158,42 @@
...
+                <constraints nullable="false" />
+            </column>
+            <column name="description" type="text" />
+            <column name="testid" type="integer">
+                <constraints nullable="false" />
+            </column>
+            <column name="run_id" type="integer">
+                <constraints nullable="false" />
+            </column>
+            <column name="owner" type="text">
```
rvansa, line 2184: I don't see owner and access mapped in the entity.

horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/DataSet.java
```diff
@@ -0,0 +1,65 @@
...
+
+   @ManyToOne(cascade = CascadeType.DETACH, fetch = FetchType.LAZY)
```
rvansa, line 63: When serializing into JSON we shouldn't send the whole Run as well, I would send just `runId`.

horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/Run.java
```diff
@@ -54,6 +57,9 @@
    @Column(columnDefinition = "boolean default false")
    public boolean trashed;
 
+   @OneToMany(mappedBy = "run", cascade = CascadeType.ALL, orphanRemoval = true)
+   public Collection<DataSet> datasets;
```
rvansa, line 61: I wonder if bi-directional mapping will serve us rather than cause trouble (not a request for change, just remark).

horreum-backend/src/main/resources/db/changeLog.xml
johnaohara: Should `DataSet` entity extend from the same `ProtectedBaseEntity` abstract class to inherit the same semantics as `Run`?  @rvansa A quick question about current `Test` and `Run` entities? Is there a reason that `Test` doe not extend `ProtectedBaseEntity`? `Test` can have multiple `TestToken` tokens registered, but `Run` can only have a single String `token`, why are the token security semantics different for `Test` and `Run`?

rvansa: The system of tokens is going to change anyway as it doesn't suit the inteded purpose very well. My current plan is that you could have a token that would grant access to multiple items, so that with one token you could grant access to both Run (or DataSet) and Test.

@whitingjr I think that in the policies you could ignore any tokens for now. That would probably mean not extending `ProtectedBaseEntity` as we would have unused `token` column.

`Test` originally extended `ProtectedBaseEntity` and used single token but I have changed that when converting to multiple tokens. I didn't want to tackle `Run` at the same time.

horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/DataSet.java
whitingjr: ok I'll remove this field.

horreum-backend/src/main/resources/db/changeLog.xml
whitingjr: Yes.

horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/DataSet.java
whitingjr: ok

horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/Run.java
whitingjr: It's too early to tell if it's going to be a problem. Let's postpone any verdict.

horreum-backend/src/main/resources/db/changeLog.xml
whitingjr: I added an intermediate super-class OwnedEntityBase to hold owner and access. ProtectedEntityBase extends that.

horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/json/DataSet.java
rvansa: This way you wouldn't send the `runId` at all. You should provide a getter annotated with `@JsonProperty`

horreum-backend/src/main/resources/db/changeLog.xml
rvansa: You are missing the policy for reading datasets:
```
CREATE POLICY dataset_select ON dataset FOR SELECT
    USING (can_view2(access, owner) OR has_role('horreum.system'));
```

rvansa (commented):
Almost ready.

rvansa:
Integrated, thanks!

rvansa:
Ouch. I should read twice before pushing the merge button - you've added the policy for SELECT but it's completely off, giving read access to uploaders and not allowing anyone else (without upload token) to  read the data! I'll fix this before the changelog gets applied anywhere.

PR #151, opened 2022-03-09
Add swagger alert for token param

dupliaka:
To avoid misunderstandings in token I suggest adding a description to the swagger


rvansa:
Integrated, thanks!

PR #174, opened 2022-04-11
Save score units on the test level and represent it on the changes

dupliaka:
Feature request on unit labeling 
Looks like:
image

Need to update container to include util to the db

### DIFF
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/AlertingService.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/AlertingService.java
@@ -87,6 +87,7 @@ public interface AlertingService {
       public int testId;
       public String uid;
       public String url;
+      public String unit;
       public List<PanelInfo> panels = new ArrayList<>();
    }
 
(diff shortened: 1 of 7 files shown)


rvansa:
This should definitely not be a property of the Test; you could have it as a property of the Variable, though.

Visually I would put it vertical and vertically centered rather than angled in top left.

dupliaka:
1. In case you have different units you just do not define it in the test and define it as a part of the variable name. I dont see any profit in adding it to variables since you can then make it a  part of the name. 
2. If I put it vertically then it overlaps with the titles of the scale. I tried to put it on the top without any angling but it either overlaps the scale number or it go behind the canvas.

dupliaka:
Maybe then put the label on the mid as at reports

rvansa:
1. So you want to add a "default unit" for all variable names, rather than repeating it for every variable? Hmm, okay, but the it should be still defined in the "Change Detection" tab, and probably hosted in another table. Putting it straight to Test seems very ad-hoc, it's quite unclear what should it relate to. Should that be automatically added as unit to every View component, to Reports etc.?
2. I guess in that case you need to add some offset left from the scale.

rvansa:
1. Yes, if you really find it useful you can put it there. My preference would be having the unit individually for each variable, but I won't object to that.
2. IDK TBH - Recharts are try-and-see for me. I think that `insideLeft` is the right one, but somehow you have to make space in the parent. Maybe put the Label right into the YAxis like here: https://recharts.org/en-US/api/Label

dupliaka:
Ok, lets do one of this: resolve conflicts and merge it as it is or just close

rvansa:
I strongly disagree with adding `unit` to `Test`, therefore closing.

PR #176, opened 2022-04-22
Add relative change

dupliaka:
This change adds relative change feature to Reports

Baseline setup is getting stored in report edit window at report config tabel
When report gets generated the runs related to baseline marks as baseline (stores in report data table)
The calculation is made in UI

I did backend and uI changes + db migration

I did not wrote tests and hide baseline if it was not declared (Need some help)

image
[sqlFile.zip](https://github.com/Hyperfoil/Horreum/files/8541078/sqlFile.zip)

### DIFF
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/report/TableReport.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/report/TableReport.java
@@ -72,6 +72,8 @@ public class TableReport extends PanacheEntityBase {
       @NotNull
       public ArrayNode values;
 
+      public boolean baseline;
+
       @Override
       public String toString() {
          return "TableReport.Data{" +
@@ -79,6 +81,7 @@ public class TableReport extends PanacheEntityBase {
                ", category='" + category + '\'' +
                ", series='" + series + '\'' +
                ", label='" + scale + '\'' +
+               ", label='" + baseline + '\'' +
                ", values=" + values +
                '}';
       }
--- a/horreum-backend/src/main/resources/db/changeLog.xml
+++ b/horreum-backend/src/main/resources/db/changeLog.xml
@@ -696,7 +696,7 @@
                 ));
         </sql>
     </changeSet>
-    
+
     <changeSet id="5" author="rvansa">
         <createTable tableName="userinfo">
             <column name="username" type="text">
@@ -1833,7 +1833,7 @@
                 USING (has_role2((SELECT test.owner FROM test JOIN tablereportconfig trc ON test.id = trc.testid JOIN tablereport tr ON tr.config_id = trc.id WHERE tr.id = report_id), 'tester'));
         </sql>
     </changeSet>
-    
+
     <changeSet id="34" author="rvansa">
         <createTable tableName="reportcomment">
             <column name="id" type="integer">
@@ -2437,7 +2437,7 @@
                 USING (exists(SELECT 1 FROM label WHERE label.id = label_id AND has_role2(label.owner, 'tester')));
         </sql>
     </changeSet>
-    
+
     <changeSet id="60" author="rvansa">
         <createTable tableName="label_values">
             <column name="dataset_id" type="integer">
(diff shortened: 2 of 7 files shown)


horreum-api/src/main/java/io/hyperfoil/tools/horreum/entity/report/TableReport.java
```diff
@@ -72,13 +72,16 @@ public static class Data {
...
+      public boolean baseline;
+
       @Override
       public String toString() {
          return "TableReport.Data{" +
                "datasetId=" + datasetId +
                ", category='" + category + '\'' +
                ", series='" + series + '\'' +
                ", label='" + scale + '\'' +
+               ", label='" + baseline + '\'' +
```
rvansa, line 84: please change `label=` to `scale=` and use `baseline=` for yours.

horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -3102,6 +3102,15 @@
             CREATE TRIGGER ds_after_schema_update AFTER INSERT OR UPDATE OF uri ON schema FOR EACH ROW EXECUTE FUNCTION ds_after_schema_update_func();
         </sql>
     </changeSet>
+    <changeSet id="70" author="dupliaka">
+        <addColumn tableName="tablereport_data">
+            <column name="baseline" type="boolean"/>
```
rvansa, line 3107: For boolean it would be better to have a non-null constraint. In changeset you need to do that by defining it with default `false` as well and then immediately `dropDefault`.

webapp/src/domain/reports/TableReportConfigPage.tsx
```diff
@@ -298,6 +299,59 @@ export default function TableReportConfigPage() {
...
                         </FormSection>
+
+                        <FormSection
+                            title={
+                                <>
+                                    Baseline
+                                    <Popover
+                                        headerContent="Baseline labels and function"
+                                        bodyContent={
+                                            <div>
```
rvansa, line 310: It would be useful to mention here that setting these is optional, and allows you to select another set of DataSets.

rvansa (commented):
I wonder how this works: `getFinalDatasetIds` makes sure that for every combo of (category, series, scale) we get a single (most recent by timestamps) result; if you're not taking baseline as a parameter for this combo I think you'll always get only one of baseline/non-baseline result.

dupliaka:
> I wonder how this works: `getFinalDatasetIds` makes sure that for every combo of (category, series, scale) we get a single (most recent by timestamps) result; if you're not taking baseline as a parameter for this combo I think you'll always get only one of baseline/non-baseline result.

Can you please make an example?

rvansa:
What example? If you have two categories, say `JVM` and `native` (and you're not going to compare them but have them both in the same report), series can be e.g. `v1.9` and `v2.0` and scale is 1 (CPU), 2 (CPU), 4 (CPU)... Then you have 12 possible combinations, and if every combination has a matching DataSet `getFinalDatasetIds` will give you 12 datasets to operate on. However if you use baseline, you should pick 12 datasets for the 'updated' and 12 for 'baseline' so that you can calculate the respective relative difference.
Or please show your test data.

dupliaka:
> What example? If you have two categories, say `JVM` and `native` (and you're not going to compare them but have them both in the same report), series can be e.g. `v1.9` and `v2.0` and scale is 1 (CPU), 2 (CPU), 4 (CPU)... Then you have 12 possible combinations, and if every combination has a matching DataSet `getFinalDatasetIds` will give you 12 datasets to operate on. However if you use baseline, you should pick 12 datasets for the 'updated' and 12 for 'baseline' so that you can calculate the respective relative difference. Or please show your test data.

For the test data you can use this
[sqlFile.zip](https://github.com/Hyperfoil/Horreum/files/8541078/sqlFile.zip)

`PGPASSWORD=secret psql -h localhost -U dbadmin horreum -f sqlFile.zip`

dupliaka:
https://github.com/Hyperfoil/Horreum/pull/195

PR #185, opened 2022-04-28
Move Jackson JSON converters to horreum-api module

### LINKED ISSUE #184: Can not add dependency to horreum-api without horreum-backend
`horreum-backend` module contains JSON Jackson converters used for enitites in `horreum-api`

Users wishing to write an application to register as a webhook need to add a dependency on both `horreum-api` and `horreum-backend` modules to be able to use entities defined in `horreum-api` as objects in their webhook application.

e.g. for `io.hyperfoil.tools.horreum.entity.json.Run` the data is marshalled by;

```
   @NotNull
   @Type(type = "io.hyperfoil.tools.horreum.entity.converter.JsonUserType")
   public JsonNode data;
```
`io.hyperfoil.tools.horreum.entity.converter.JsonUserType` is currently in `horreum-backend`

johnaohara:
This PR moves the Jackson JSON converters to the `horreum-api` module to allow users writing downstream webhooks to simply depend on `horreum-api`

fixes #184


rvansa:
Thanks, integrated!

PR #192, opened 2022-05-10
Bump horreum client to resteasy 4.7.5.Final

johnaohara:
Align Horreum client resteasy client version to Quarkus resteasy client version.


johnaohara:
Looks like this change was merged directly into main branch

PR #214, opened 2022-06-09
Pass through property values.

whitingjr:
Pass through password properties to testcontainers environment for starting Horreum container correctly.


rvansa (approved):
I am not particularly happy about another linking to a different module, but I am going to integrate as-is and deal with it later on.

PR #217, opened 2022-06-14
Split out integration test to separate project module.

### LINKED ISSUE #216: Split integration tests into separate module.
The integration tests should be split into a separate module.

 A desirable outcome is to ensure infrastructure containers to start only once for all the test-cases.

whitingjr:
This PR resolves #216 issue to add an enhancement.


horreum-integration/src/test/java/io/hyperfoil/tools/HorreumTestBase.java
```diff
@@ -42,7 +46,10 @@
 
 import io.hyperfoil.tools.horreum.entity.json.Test;
 
-public class HorreumTestBase {
+import io.restassured.RestAssured;
+import io.restassured.specification.RequestSpecification;
+
+public class HorreumTestBase implements BeforeAllCallback, AfterAllCallback, ExtensionContext.Store.CloseableResource  {
```
rvansa, line 52: Looks like you made this an extension, it is not used as 'test base' anymore - the name should reflect that.

And about `HorreumTestClientBase` - the original `HorreumTestBase` class is used for Jenkins plugins tests as well, and therefore it should create the infrastructure & start Horreum but **not** initialize anything specific to the resteasy client. Please layer the classes for reuse in the Jenkins plugin tests. And make sure that any module that includes the base class is publishing test-jars.

whitingjr: > Looks like you made this an extension, it is not used as 'test base' anymore - the name should reflect that.
 Ok the name will be updated.

> And about HorreumTestClientBase - the original HorreumTestBase class is used for Jenkins plugins tests as well, and therefore it should create the infrastructure & start Horreum but not initialize anything specific to the resteasy client. 
I have had no end of strife (InitialisationError errors) initializing the Resteasy client in multiple places. Doing it instead once and sharing the built instance works.

> Please layer the classes for reuse in the Jenkins plugin tests. 

 I have created a type hierarchy here so that a subsequent integration test (to be added in another commit) avoids creating the HorreumClient.

> And make sure that any module that includes the base class is publishing test-jars.
ok I will add that.

.github/workflows/main.yml
```diff
@@ -29,7 +30,9 @@ jobs:
     - name: Maven Version
       run: mvn --version
     - name: Build with Maven
-      run: mvn clean -B package --file pom.xml
+      run: mvn clean package -B --file pom.xml -Dnpm.skip -DskipTests
```
rvansa, line 33: Why do you skip NPM build? If there's a problem with that the check won't figure out.

horreum-backend/pom.xml
```diff
@@ -179,6 +179,16 @@
     </dependencies>
 
     <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
```
rvansa, line 185: Plugin version 3.2.0 is already set in parent POM.

horreum-integration/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+    <pluginManagement>
```
rvansa, line 86: incorrect indentation

horreum-integration/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+        <plugins>
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
```
rvansa, line 89: Why is this in `pluginManagement` section rather than in `plugins`? Will it run at all?

horreum-integration/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+                                </propertyfile>
+                            </target>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>3.2.0</version>
```
rvansa, line 114: Same as above, should be in parent POM.

horreum-integration/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+        <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+                <systemPropertyVariables>
+                    <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                    <maven.home>${maven.home}</maven.home>
+                    <horreum.commit.id>${buildNumber}</horreum.commit.id>
+                </systemPropertyVariables>
+                <additionalClasspathElements>
```
rvansa, line 139: What is this for? Use proper dependencies, please.

horreum-integration/pom.xml
```diff
@@ -0,0 +1,185 @@
...
+            </includes>
+        </testResource>
+        <testResource>
+            <directory>../horreum-backend/src/main/resources</directory>
+            <targetPath>docker-compose</targetPath>
+            <includes>
+                <include>keycloak-horreum.json</include>
+            </includes>
+        </testResource>
+        <testResource>
```
rvansa, line 177: Just add another include above?
I also wonder, since referencing path another module confuses IntelliJ if it wouldn't be better to just replace it with a symlink...

horreum-integration/src/test/java/io/hyperfoil/tools/HorreumClientTest.java
```diff
@@ -13,9 +13,11 @@
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.extension.ExtendWith;
 
+@ExtendWith({HorreumTestExtension.class})
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
-public class HorreumClientTest extends HorreumTestBase {
+public class HorreumClientTest extends HorreumTestClientExtension {
```
rvansa, line 20: Should it extend `HorreumTestClientExtension`, or just use `@ExtendWith(HorreumTestClientExtension.class)`? I am a bit confused here...

horreum-integration/src/test/java/io/hyperfoil/tools/HorreumTestExtension.java
```diff
@@ -319,4 +340,9 @@ public TestContainer(String composeFile, Map<String, String> envVariables) {
             withEnv(envVariables);
         }
     }
+
+    public static RequestSpecification bareRequest(){
```
rvansa, line 344: Where is this used? Integration tests should use the available clients directly, not through RestAssured.

.github/workflows/main.yml
whitingjr: To speed up the build. I can remove it though.

horreum-backend/pom.xml
whitingjr: Good I can remove this.

horreum-integration/pom.xml
whitingjr: thanks for spotting that.

horreum-integration/pom.xml
rvansa: keycloak-horreum.json is in backend resources only to become part of the image that's reused as Keycloak init-container in Openshift; we can do better.

horreum-integration/pom.xml
whitingjr: Much of the pom file was copied from horreum-client module. Will move the configuration to the correct place.

horreum-integration/pom.xml
whitingjr: Again, copied from horreum-client. Will be removed.

horreum-integration/src/test/java/io/hyperfoil/tools/HorreumTestExtension.java
whitingjr: This is used in a subsequent PR. Will remove it from this one.

horreum-integration/src/test/java/io/hyperfoil/tools/HorreumClientTest.java
whitingjr: In the desperation to get something working I munged together composition and inheritance. I'll separate it out for clarity. Thanks for spotting this.

horreum-integration/pom.xml
whitingjr: Will remove that. Thanks.

whitingjr:
@rvansa looks like this is failing due to https://github.com/Hyperfoil/Horreum/blob/master/horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TokenAuthTest.java#L25

rvansa:
@whitingjr Meh, flaky tests... We'll have to deal with it in another issue. I've merged your other PR - please rebase & squash the commits, and I'll get this in.

rvansa:
Integrated, thanks!

PR #306, opened 2022-11-22
podman: Enable socket environment

jesperpedersen:


README.md
```diff
@@ -34,15 +34,23 @@ and after a few moments everything should be up and ready. The script will creat
...
-dnf install -y podman podman-plugins podman-compose
+dnf install -y podman podman-plugins podman-compose podman-docker
+```
+
+Enable socket environment:
+
+``` bash
+systemctl --user enable --now podman.socket
+export DOCKER_HOST=unix:///run/user/${UID}/podman/podman.sock
+export TESTCONTAINERS_RYUK_DISABLED=true
```
rvansa, line 45: You can disable Ryuk in some properties file in the test JAR, so that user does not have set it here.

README.md
```diff
@@ -34,15 +34,23 @@ and after a few moments everything should be up and ready. The script will creat
 Install:
 
 ``` bash
-dnf install -y podman podman-plugins podman-compose
+dnf install -y podman podman-plugins podman-compose podman-docker
+```
+
+Enable socket environment:
```
rvansa, line 40: The whole paragraph is missing context - why you need to enable socket, is this needed for running in podman or just compiling the project etc. (I know the answers, but please explain things in the README).

jesperpedersen: I added a note... The README is too long as is.

It will help - a bit - once we have a newer npm environment that supports OpenSSL 1.1.x and 3 out of the box, and doesn't need the hack

PR #317, opened 2022-11-28
Remove Maven wrapper scripts

jesperpedersen:


johnaohara:
If we are removing the helper scripts from the Quarkus project, we need to document that devs need the maven installed with the correct version as a prerequisite to building the project; otherwise, the Quarkus maven plugins will not function.

jesperpedersen:
Its in "Prerequisites"

jesperpedersen:
We need to have a clear directory directory layout such that it is easy to figure out where your contribution goes

rvansa:
What does clear directory layout have to do with Maven wrapper? Having that is a standard thing for any Quarkus project.

rvansa:
To digress a bit, why are Keycloak, Grafana and PostgreSQL in *Prerequisites*?  You don't need to have that installed locally. On the other hand, you need either Docker or Podman.
Sure, we depend on these but they're transparently managed via images. So I wouldn't list them (we don't list Quarkus either and we depend on that...)

jesperpedersen:
They are prerequisites since the test suite won't run without them - docker or podman is a way to get them running locally.

rvansa:
In my view *Prerequisites* are the thing you need to do/get manually. Testsuite will run just fine even if you never touch them. I see these as *Dependencies*.

Issue #348, opened 2023-01-25
Typo in https://raw.githubusercontent.com/Hyperfoil/Horreum/0.7/infra/start.sh


vitorz:
**Describe the bug**

In the [script](https://raw.githubusercontent.com/Hyperfoil/Horreum/0.7/infra/start.sh) to run locally an Horreum instance with podman, there is a typo regarding the version 0.7 of the product.
In fact in the script the old version 0.6 is reported instead of the right one: 0.7

**Version**

0.7

jesperpedersen:
Do you want to provide a patch for this ?

We likely need something else in the future in order to make sure that the version is always up-to-date.

johnaohara:
@jesperpedersen +1 I think we need to review the release process, to automate updating references such as this during the release

vitorz:
@jesperpedersen  or @johnaohara  what do you think about 2 PR one for branch 0.7.x and one for master? Feel free to suggest a better way.

jesperpedersen:
@vitorz For now, just submit a PR for master using `0.7` as the content. I can cherry-pick it to the `0.7.x` branch.

We can do a discussion thread on what going forward

vitorz:
ok, done: #350

jesperpedersen:
Fixed in 9e1e32655a87f9d9db5904190725de050f3f5ec0

PR #350, opened 2023-01-25
version changed to more recent release in start.sh

vitorz:
#In `start.sh` changed the version value from `0.6` to `0.7,` which is the more recent one. Issue: #348


jesperpedersen:
Merged.

PR #368, opened 2023-02-08
Increase timeout for clearing data for tests. 

johnaohara:
Verify data is cleared ~10 times insteead of 1


jesperpedersen:
Please, squash your commits

johnaohara:
I will do when I am finished

johnaohara:
I am going to close this for now, it has turned into a bigger refactor than originally anticipated as the scope has grown, but the end result  will be a much simpler integration testsuite

PR #392, opened 2023-02-17
remove columns in All.tsx file

shivam-sharma7:
- remove Id column
- Move Owner column after run
- remove Access column


jesperpedersen:
We need to keep the Access padlock in the Owner column, so that needs to be implemented.

You should also look at AllSchema.tsx

shivam-sharma7:
Access padlock any idea?

jesperpedersen:
Thanks for working on this !

For now, the work has been completed through #402 and #404 have been filed for the remaining work within this area.

Issue #393, opened 2023-02-17
Listing runs by Schema causes exception due to PG type mapping missing for NativeQuery


whitingjr:
**Describe the bug**

When trying a Extractor expression for a Schema that has run_schemas an exception is thrown.

> Failed to fetch runs by Schema URI.
> 
> {
>   "details": "Error id 271a1428-5bb0-4959-a5ef-53092a053051-73",
>   "stack": ""
> }
> 
> 	2023-02-14 16:40:17,848 ERROR [org.jbo.res.rea.com.cor.AbstractResteasyReactiveContext] (executor-thread-3466) Request failed: javax.persistence.PersistenceException: org.hibernate.MappingException: No Dialect mapping for JDBC type: 1111
> 	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:154)
> 	at org.hibernate.query.internal.AbstractProducedQuery.list(AbstractProducedQuery.java:1626)
> 	at org.hibernate.query.Query.getResultList(Query.java:165)
> 	at io.hyperfoil.tools.horreum.svc.RunServiceImpl.listBySchema(RunServiceImpl.java:794)
> 	at io.hyperfoil.tools.horreum.svc.RunServiceImpl_Subclass.listBySchema$$superforward1(Unknown Source)
> 	at io.hyperfoil.tools.horreum.svc.RunServiceImpl_Subclass$$function$$6.apply(Unknown Source)
> 	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:54)
> 	at io.hyperfoil.tools.horreum.server.TokenInterceptor.wrap(TokenInterceptor.java:49)
> 	at io.hyperfoil.tools.horreum.server.TokenInterceptor_Bean.intercept(Unknown Source)
> 	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
> 	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:50)
> 	at io.hyperfoil.tools.horreum.server.RolesInterceptor.intercept(RolesInterceptor.java:70)
> 	at io.hyperfoil.tools.horreum.server.RolesInterceptor_Bean.intercept(Unknown Source)
> 	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
[...]

PR #394, opened 2023-02-17
#393-Add column type info to NativeQuery when listing runs by Schema

whitingjr:
This PR looks to solve #393. By adding type information to the NativeQuery the resultset can be processed without causing an exception. Reported due to an unrecognized type 1111. A Types.OTHER type.


jesperpedersen:
Please, remove unrelated changes, and indent the new code correctly. Then, squash and force push

whitingjr:
The build is failing due to the environment property NODE_OPTIONS not being set.

jesperpedersen:
Merged.

PR #395, opened 2023-02-20
Add Null Pointer check to recalcdataset

johnaohara:
If a user clicks "Recalculate" datasets multiple times, there is a race condition where a dataset might have already been deleted and no longer exists. In this situation `recalculateForDataset` throws a NPE. If a dataset does not exist it should be logged instead


jesperpedersen:
Merged - after changing the log level.

PR #407, opened 2023-03-03
added openapi annotation for the response of RunService.addRunFromData

vitorz:
Adding this openapi annotation for the response of `RunService.addRunFromData`, the generated yaml file with the api definition is more correct, allowing to produce better client for the run service.


jesperpedersen:
Rebased, and merged.

PR #432, opened 2023-03-15
Do not null out test from reportConfig - Fixes #431

### LINKED ISSUE #431: Reports are not linked to Tests
**Describe the bug**

When reports configurations are persisted to disk, the report mapping is removed. This causes errors in the UI where reports are not associated with a test;

Screenshot from 2023-03-15 21-12-04

This issue was introduced in this commit: https://github.com/Hyperfoil/Horreum/commit/c207310a  Unfortunately we do not have details of the original issue

johnaohara:
Fixes #431


shivam-sharma7 (approved):
LGTM 👌

jesperpedersen:
I rewrote it a bit, and ref'd the issue in the commit message. Rebased, and merged.

PR #437, opened 2023-03-16
Upgrade keycloak

shivam-sharma7:
- upgrade keycloak version 
close: #435


jesperpedersen:
You need more changes - `git grep keycloak`

johnaohara:
@jesperpedersen is issue #435 to upgrade keycloak across the entire stack? If that is the case, we need to upgrade the projects that depend on horreum (e.g. jenkins plugin etc) as well to ensure API compatibility

jesperpedersen:
And, we have to keep in mind the new management API for Keycloak

jesperpedersen:
> You need more changes - `git grep keycloak`

Still true

PR #444, opened 2023-03-21
[#443] upgarade: netty from 1.1.87.final to 4.1.90.final

shivam-sharma7:
upgarade: netty from 1.1.87 to 4.1.90


jesperpedersen:
Rabased, and merged.

PR #446, opened 2023-03-22
Updated dev env to work on Windows with Docker Desktop and WSL2

johnaohara:


infra/docker-compose.yml
```diff
@@ -5,43 +5,43 @@ services:
...
   db-init:
     depends_on:
     - postgres
     image: docker.io/postgres:13
     restart: on-failure
     env_file: postgres.env
     environment:
-    # DB-init resolves localhost to ::1 rather than 127.0.0.1
-    - PGHOST=127.0.0.1
+    - PGHOST=172.17.0.1
```
jesperpedersen, line 21: How does these changes affect a Linux/Docker based env ?

jesperpedersen (changes requested):

infra/docker-compose.yml
johnaohara: We were relying on host networking, which does not work in Windows. so have switch to the default network mode.  db-init needs to connect to the postgres database, which is no longer localhost from inside a container and we have to connect of the docker bridge.  It should work exactly the same on Linux.  I am verifying now

johnaohara: @jesperpedersen , I have verified that it works as expected on Linux

jesperpedersen:
Rebased, and merged.

Issue #445, opened 2023-03-21
Integrate Jest


jesperpedersen:
Use https://jestjs.io/ as the UI testing framework

johnaohara:
Closing as a duplicate of #901

PR #451, opened 2023-03-23
[#445] Add jest

shivam-sharma7:
close: #445


jesperpedersen:
We need some documentation (`doc/TESTING.md`), and an example on how to write a test case - f.ex. a test case that verifies the front page

PR #453, opened 2023-03-23
Enforce that authenticated Managers and Admin users should only be al…

johnaohara:
…lowed to add new users. Do no display admn panel for non-admin users


jesperpedersen:
Rebased, and merged.

PR #482, opened 2023-04-24
Ensure rendering of fingerprint options is consistent across all test…

### LINKED ISSUE #481: Inconsistent rendering of Changes fingerprint filters
In the "Changes" tab, where fingerprints have been defined in Change detection for a test, rendering of the fitlering drop-down changes depending on how many filter options there are available. If there are less than 16 options, a single dropdown list is displayed. More than 16, then the fingerprint options are split into seperate selection boxes for the user to choose.  

This introduces an inconsistent UI experience, where the UI differs between tests, or can suddenly change if the threshold of 16 options is crossed.

Less than 16;

Screenshot from 2023-04-24 17-33-45

More than 16;

Screenshot from 2023-04-24 17-34-22

johnaohara:
…s - Fixes #481

### DIFF
--- a/webapp/src/domain/alerting/Changes.tsx
+++ b/webapp/src/domain/alerting/Changes.tsx
@@ -275,6 +275,7 @@ export default function Changes() {
                                         selection={selectedFingerprint}
                                         onSelect={setSelectedFingerprint}
                                         source={fingerprintSource}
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

PR #496, opened 2023-04-27
add node.js CI

shivam-sharma7:

- [x] My change requires changes to the documentation.
- [x] I have updated the documentation accordingly.


jesperpedersen:
Why ?

shivam-sharma7:
> Why ?

for automate the build, test, and deployment process for Node.js

johnaohara:
We do not use node.js for anything for production. Node.js is used for development (which is handled by Quinoa) and build time compilation (part of current CI tasks). There is no server-side javascript/typescript in this project

shivam-sharma7:
> We do not use node.js for anything for production. Node.js is used for development (which is handled by Quinoa) and build time compilation (part of current CI tasks). There is no server-side javascript/typescript in this project

okay closed now

shivam-sharma7:
> We do not use node.js for anything for production. Node.js is used for development (which is handled by Quinoa) and build time compilation (part of current CI tasks). There is no server-side javascript/typescript in this project

But I want to add for front-end like someone install new dep using `npm install` in webapp dir

johnaohara:
> But I want to add for front-end like someone install new dep using `npm install` in webapp dir

This should be handled by Quinoa.

If you have a concrete example of a new dependency to install, we can cover how to do this in documentation

PR #510, opened 2023-05-03
Add: AllSchema.tsx Pagination

shivam-sharma7:
## Screenshort

image

 

### DIFF
--- a/webapp/src/domain/schemas/AllSchema.tsx
+++ b/webapp/src/domain/schemas/AllSchema.tsx
@@ -1,11 +1,11 @@
 import { useMemo, useEffect, useState } from "react"
 import { useSelector } from "react-redux"
 import { useDispatch } from "react-redux"
-import { Card, CardHeader, CardBody, PageSection } from "@patternfly/react-core"
+import { Card, CardHeader, CardFooter, CardBody, PageSection, Pagination } from "@patternfly/react-core"
 import { NavLink } from "react-router-dom"
 
 import * as actions from "./actions"
-import * as selectors from "./selectors"
+import { alertAction } from "../../alerts"
 import { useTester, teamsSelector, teamToName } from "../../auth"
 import { noop } from "../../utils"
 import Table from "../../components/Table"
@@ -14,7 +14,7 @@ import ActionMenu, { useShareLink, useChangeAccess, useDelete } from "../../comp
 import ButtonLink from "../../components/ButtonLink"
 import { CellProps, Column } from "react-table"
 import { SchemaDispatch } from "./reducers"
-import { Access, Schema } from "../../api"
+import Api, { Access, SortDirection, SchemaQueryResult, Schema } from "../../api"
 import SchemaImportButton from "./SchemaImportButton"
 
 type C = CellProps<Schema>
@@ -22,6 +22,25 @@ type C = CellProps<Schema>
 export default function AllSchema() {
     document.title = "Schemas | Horreum"
     const dispatch = useDispatch<SchemaDispatch>()
+
+    const [page, setPage] = useState(1)
+    const [perPage, setPerPage] = useState(20)
+    const [direction] = useState<SortDirection>("Ascending")
+    const pagination = useMemo(() => ({ page, perPage, direction }), [page, perPage, direction])
+    const [schemas, setSchemas] = useState<SchemaQueryResult>()
+    const [loading, setLoading] = useState(false)
+
+    useEffect(() => {
+        setLoading(true)
+        Api.schemaServiceList(
+             SortDirection.Ascending,
+             pagination.perPage,
+              pagination.page
+             )
+            .then(setSchemas)
+            .catch(error => dispatch(alertAction("FETCH_REPORTS", "Failed to fetch reports", error)))
+            .finally(() => setLoading(false))
+    }, [pagination,  dispatch])
 
     const columns: Column<Schema>[] = useMemo(
         () => [
@@ -83,7 +102,6 @@ export default function AllSchema() {
         [dispatch]
     )
     const [reloadCounter, setReloadCounter] = useState(0)
-    const list = useSelector(selectors.all)
     const teams = useSelector(teamsSelector)
     useEffect(() => {
         dispatch(actions.all()).catch(noop)
@@ -96,18 +114,27 @@ export default function AllSchema() {
                     <CardHeader>
                         <ButtonLink to="/schema/_new">New Schema</ButtonLink>
                         <SchemaImportButton
-                            schemas={list || []}
+                            schemas={schemas?.schemas || []}
                             onImported={() => setReloadCounter(reloadCounter + 1)}
                         />
                     </CardHeader>
                 )}
                 <CardBody style={{ overflowX: "auto" }}>
                     <Table columns={columns}
-                    data={list || []}
+                    data={schemas?.schemas || []}
                     sortBy={[{ id: "name", desc: false }]}
-                     
+                    isLoading={loading}
                     />
                 </CardBody>
+                <CardFooter style={{ textAlign: "right" }}>
+                    <Pagination
+                        itemCount={schemas?.count || 0}
+                        perPage={perPage}
+                        page={page}
+                        onSetPage={(e, p) => setPage(p)}
+                        onPerPageSelect={(e, pp) => setPerPage(pp)}
+                    />
+                </CardFooter>
             </Card>
         </PageSection>
     )
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
@@ -187,15 +187,15 @@ public class SchemaServiceImpl implements SchemaService {
    @PermitAll
    @WithRoles
    @Override
-   public List<Schema> list(Integer limit, Integer page, String sort, SortDirection direction) {
+   public SchemaQueryResult list(Integer limit, Integer page, String sort, SortDirection direction) {
       if (sort == null || sort.isEmpty()) {
          sort = "name";
       }
       Sort.Direction sortDirection = direction == null ? null : Sort.Direction.valueOf(direction.name());
       if (limit != null && page != null) {
-         return Schema.findAll(Sort.by(sort).direction(sortDirection)).page(Page.of(page, limit)).list();
+         return new SchemaQueryResult(Schema.findAll(Sort.by(sort).direction(sortDirection)).page(Page.of(page, limit)).list(), Schema.count());
       } else {
-         return Schema.listAll(Sort.by(sort).direction(sortDirection));
+         return new SchemaQueryResult(Schema.listAll(Sort.by(sort).direction(sortDirection)), Schema.count());
       }
    }
 
--- a/horreum-model/src/main/java/io/hyperfoil/tools/horreum/api/SchemaService.java
+++ b/horreum-model/src/main/java/io/hyperfoil/tools/horreum/api/SchemaService.java
@@ -21,7 +21,6 @@ import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.networknt.schema.ValidationMessage;
 
 import io.hyperfoil.tools.horreum.entity.json.Label;
 import io.hyperfoil.tools.horreum.entity.json.Schema;
@@ -44,10 +43,10 @@ public interface SchemaService {
    Integer add(Schema schema);
 
    @GET
-   List<Schema> list(@QueryParam("limit") Integer limit,
-                     @QueryParam("page") Integer page,
-                     @QueryParam("sort") String sort,
-                     @QueryParam("direction") @DefaultValue("Ascending") SortDirection direction);
+   SchemaQueryResult list(@QueryParam("limit") Integer limit,
+                          @QueryParam("page") Integer page,
+                          @QueryParam("sort") String sort,
+                          @QueryParam("direction") @DefaultValue("Ascending") SortDirection direction);
 
 
    @GET
@@ -131,6 +130,18 @@ public interface SchemaService {
    @Path("import")
    @Consumes(MediaType.APPLICATION_JSON)
    void importSchema(JsonNode config);
+
+   class SchemaQueryResult {
+      @NotNull
+      public List<Schema> schemas;
+      @JsonProperty(required = true)
+      public long count;
+
+      public SchemaQueryResult(List<Schema> schemas, long count) {
+         this.schemas = schemas;
+         this.count = count;
+      }
+   }
 
    @org.eclipse.microprofile.openapi.annotations.media.Schema(anyOf = {
          LabelInFingerprint.class, LabelInRule.class, LabelInReport.class, LabelInVariable.class, LabelInView.class
--- a/webapp/src/domain/schemas/actions.ts
+++ b/webapp/src/domain/schemas/actions.ts
@@ -14,7 +14,7 @@ const loaded = (schema: Schema | Schema[]): LoadedAction => ({
 export function getById(id: number) {
     return (dispatch: Dispatch<LoadedAction | AddAlertAction>) =>
         Api.schemaServiceGetSchema(id).then(
-            response => dispatch(loaded(response)),
+            response => dispatch(loaded(response.schema)),
             error => {
                 dispatch(loaded([]))
                 return dispatchError(dispatch, error, "GET_SCHEMA", "Failed to fetch schema")
@@ -36,7 +36,7 @@ export function add(payload: Schema) {
 export function all() {
     return (dispatch: Dispatch<LoadedAction | AddAlertAction>) =>
         Api.schemaServiceList().then(
-            response => dispatch(loaded(response)),
+            response => dispatch(loaded(response.schemas)),
             error => {
                 dispatch(loaded([]))
                 return dispatchError(dispatch, error, "LIST_SCHEMAS", "Failed to list schemas")


shivam-sharma7:
@jesperpedersen

johnaohara:
@shivam-sharma7 thank you for tackling this issue.

We need the pagination feature to apply the Schema dataset as well, so that Schemas are displayed according to the user selection in the UI

shivam-sharma7:
> We need the pagination feature to apply the Schema dataset as well, so that Schemas are displayed according to the user selection in the UI

I dont't understand. any idea?

johnaohara:
Current the PR adds the pagination function to the view, the remaining step is to apply the user selection to the `Schemas` dataset, i.e. paginate the `Schema` dataset based on the user input.

E.g. for Reports we have;

https://github.com/Hyperfoil/Horreum/blob/master/webapp/src/domain/reports/Reports.tsx#L59

For test Datasets we have:

https://github.com/Hyperfoil/Horreum/blob/master/webapp/src/domain/runs/TestDatasets.tsx#L147

shivam-sharma7:
> We need the pagination feature to apply the Schema dataset as well,  


Schema dataset it mean it come from `datasetServiceListBySchema` api like what already done in https://github.com/Hyperfoil/Horreum/blob/master/webapp/src/domain/reports/Reports.tsx#L59

johnaohara:
Yes, similar to how it is implemented in `Reports.tsx`, however the api call will be `Api.schemaServiceList()` with the pagination details passed as parameters.

It also looks like we are caching `Shemas` in a global state: https://github.com/Hyperfoil/Horreum/blob/master/webapp/src/domain/schemas/AllSchema.tsx#L89

We will need to decouple the UI from this global cache, otherwise we would need to create the pagination logic in the UI code.

jesperpedersen:
Please, create an issue per page

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -16,13 +16,17 @@ import { CellProps, Column } from "react-table"
 import { SchemaDispatch } from "./reducers"
 import { Access, Schema } from "../../api"
 import SchemaImportButton from "./SchemaImportButton"
+// import Schema from "./Schema"
```
jesperpedersen, line 19: We can remove this then

jesperpedersen (changes requested):

jesperpedersen:
@shivam-sharma7 Consider @johnaohara comments - we may want to bundle multiple page changes into a single pull request in order to make the pagination work correctly

webapp/src/domain/schemas/AllSchema.tsx
shivam-sharma7: yes

shivam-sharma7:
@johnaohara for set tableReport we have `<AllTableReports> component `https://github.com/Hyperfoil/Horreum/blob/0c1147ff9d591dcfa655aeed09052e0694f1d502/webapp/src/domain/reports/Reports.tsx#L52

Then, what would be for schema?

johnaohara:
For Schema the typescript interface is `Schema`.

This is interface is auto-generated by the Openapi generator, you should be able to import it by;

`import Api, { Schema } from "../../api"`

shivam-sharma7:
> 

But problem is when i try to use  `const [Schema, setSchema] = useState<Schema>();` Then i got following error.


Property 'count' does not exist on type 'Schema'.
    137 |                 <CardFooter style={{ textAlign: "right" }}>
    138 |                  <Pagination
  > 139 |                         itemCount={Schema?.count || 0}
        |                                            ^^^^^
    140 |                         perPage={perPage}
    141 |                         page={page}
    142 |                         onSetPage={(e, p) => setPage(p)}

johnaohara:
Unfortunately, there is a discrepancy in the API for schema compared to Reports. `Api.schemaServiceList()` just returns a List (Array) whereas `Api.reportServiceGetTableReports` is returning a `AllTableReports` object that contains a `count` field.

So, instead you will need to define the state hooks as something like;

`const [schemas, setSchemas] = useState<Array<Schema>>();`

and `itemCount` would become something like 

`itemCount={schemas?.length || 0}`

please note that I am not 100% proficient in typescript and I have not tested the code above, but it should be something like that.

shivam-sharma7:
`AllTableReports` is a separate component where `count` is defined that return by `Api.reportServiceGetTableReports` I'm looking at schema components like `AllTableReports` where Allschema is present or something. `AllTableReports` is present in `src/generated/models/AllTableReports` which is returned by `Api.reportServiceGetTableReports`

hooks 

    const [tableReports, setTableReports] = useState<AllTableReports>()

and `itemCount`

`itemCount={tableReports?.count || 0}` `.count` mean property which is returned by `AllTableReports`

 

#### See the `AllTableReports` component which is present in `src/generated/models/AllTableReports`

```
import { exists, mapValues } from '../runtime';
import {
    TableReportSummary,
    TableReportSummaryFromJSON,
    TableReportSummaryFromJSONTyped,
    TableReportSummaryToJSON,
} from './TableReportSummary';

/**
 * 
 * @export
 * @interface AllTableReports
 */
export interface AllTableReports {
    /**
     * 
     * @type {Array<TableReportSummary>}
     * @memberof AllTableReports
     */
    reports: Array<TableReportSummary>;
    /**
     * 
     * @type {number}
     * @memberof AllTableReports
     */
    count: number;
}

export function AllTableReportsFromJSON(json: any): AllTableReports {
    return AllTableReportsFromJSONTyped(json, false);
}

export function AllTableReportsFromJSONTyped(json: any, ignoreDiscriminator: boolean): AllTableReports {
    if ((json === undefined) || (json === null)) {
        return json;
    }
    return {
        
        'reports': ((json['reports'] as Array<any>).map(TableReportSummaryFromJSON)),
        'count': json['count'],
    };
}

export function AllTableReportsToJSON(value?: AllTableReports | null): any {
    if (value === undefined) {
        return undefined;
    }
    if (value === null) {
        return null;
    }
    return {
        
        'reports': ((value.reports as Array<any>).map(TableReportSummaryToJSON)),
        'count': value.count,
    };

```

johnaohara:
We use openapi to auto-generate the client (i.e. typescript) api code.  The REST api's for Reports and Schema are inconsistent.

You can see with Reports, that an `AllTableReports` object is returned: https://github.com/Hyperfoil/Horreum/blob/master/horreum-model/src/main/java/io/hyperfoil/tools/horreum/api/ReportService.java#L33

For Schema, the REST endpoint is returning a `List<Schema>` object: https://github.com/Hyperfoil/Horreum/blob/master/horreum-model/src/main/java/io/hyperfoil/tools/horreum/api/SchemaService.java#L47

Therefore on the client side, a function call `Api.reportServiceGetTableReports()` will return an instance of the `AllTableReports` interface, whereas a call to `Api.schemaServiceList()` will return an `Array<Schema>`.  

With the current REST API, there is no `AllSchema` equivalent of  `AllTableReports`.

So we have 2 options in this scenario;

1) we modify the REST api to return a `AllSchema` object (which contains a `count` field) instead of `List<Schama>`
or
2) we use the existing REST api and instead of calling `.count` we use the intrinsic property of `.length` for the returned `Array<Schema>`.

I am proposing option (2) as we don't have to modify the REST API, and can use the existing auto-generated client code

shivam-sharma7:
here I got `count` https://github.com/Hyperfoil/Horreum/blob/0c1147ff9d591dcfa655aeed09052e0694f1d502/horreum-model/src/main/java/io/hyperfoil/tools/horreum/api/ReportService.java#LL74C4-L79C5

johnaohara:
@shivam-sharma7 that is option (1) that I mentioned in my previous comment.  

However, this is an breaking API change (i.e. the backend API will change in a way that breaks all clients). If we want to backport this pagination feature to the 0.7 branch, we can not change the REST API.

@jesperpedersen do you want this pagination feature back-ported to the stable branch, or are you happy for it to live in master branch for now?

shivam-sharma7:
@johnaohara I 'm updating my PR with this [method](https://github.com/Hyperfoil/Horreum/pull/510#issuecomment-1534305827) first. I think it's enough.

jesperpedersen:
We will keep this in the `master` branch only

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -14,15 +14,38 @@ import ActionMenu, { useShareLink, useChangeAccess, useDelete } from "../../comp
 import ButtonLink from "../../components/ButtonLink"
 import { CellProps, Column } from "react-table"
 import { SchemaDispatch } from "./reducers"
-import { Access, Schema } from "../../api"
+import Api, { Access, Schema,} from "../../api"
+import { alertAction } from "../../alerts"
 import SchemaImportButton from "./SchemaImportButton"
-
+
```
jesperpedersen, line 20: Extra space

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -14,15 +14,38 @@ import ActionMenu, { useShareLink, useChangeAccess, useDelete } from "../../comp
...
+     
+    const pagination:any = useMemo(() => ({ page, perPage}), [page, perPage])
+     
+    const [schemas, setSchemas] = useState<Array<Schema>>();
+    const [loading, setLoading] = useState(false)
+
+     useEffect(() => {
+        setLoading(true)
+        Api.schemaServiceList(
+
```
jesperpedersen, line 38: Extra line

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -14,15 +14,38 @@ import ActionMenu, { useShareLink, useChangeAccess, useDelete } from "../../comp
...
+            pagination.perPage,
+            pagination.page,
+
```
jesperpedersen, line 41: Extra line

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -14,15 +14,38 @@ import ActionMenu, { useShareLink, useChangeAccess, useDelete } from "../../comp
...
+     useEffect(() => {
+        setLoading(true)
+        Api.schemaServiceList(
+
+            pagination.perPage,
+            pagination.page,
+             
+        )
+            .then(setSchemas)
```
jesperpedersen, line 43: Indentation for the next 3 lines

jesperpedersen (changes requested):

webapp/src/domain/schemas/AllSchema.tsx
johnaohara, line 38: we need to add the SortDirection here, otherwise the param mapping for the api call is incorrect, something like;

```
        Api.schemaServiceList(
            SortDirection.Ascending,
            pagination.perPage,
            pagination.page,
```

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -105,9 +128,19 @@ export default function AllSchema() {
...
                     data={list || []}
                     sortBy={[{ id: "name", desc: false }]}
-                     
+                    isLoading={loading}
                     />
+                     
                 </CardBody>
+                <CardFooter style={{ textAlign: "right" }}>
+                 <Pagination
+                        itemCount={schemas?.length || 0}
```
johnaohara, line 137: I had not realised that `itemCount` was total count, so the only option for this to work correctly is for us to change the REST API

johnaohara (commented):
The other changes that are needed, I can not comment on the PR directly.

We can remove https://github.com/shivam-sharma7/Horreum/blob/3cd396e31ac010de8facb1479d9042c127af700d/webapp/src/domain/schemas/AllSchema.tsx#L109 as we want to use the list of Schemas returned from the API call

We needs to change the local `schemas` variable :  https://github.com/shivam-sharma7/Horreum/blob/3cd396e31ac010de8facb1479d9042c127af700d/webapp/src/domain/schemas/AllSchema.tsx#L122  

Same as above for : https://github.com/shivam-sharma7/Horreum/blob/3cd396e31ac010de8facb1479d9042c127af700d/webapp/src/domain/schemas/AllSchema.tsx#L129

Seeing that we need to know the total number of schemas for the pagination to work correctly, we also need to modify the REST API to return the total `count` of  schemas

webapp/src/domain/schemas/AllSchema.tsx
shivam-sharma7: like what?

shivam-sharma7:
> The other changes that are needed, I can not comment on the PR directly.
> 
> We can remove https://github.com/shivam-sharma7/Horreum/blob/3cd396e31ac010de8facb1479d9042c127af700d/webapp/src/domain/schemas/AllSchema.tsx#L109 as we want to use the list of Schemas returned from the API call
> 
> We needs to change the local `schemas` variable : https://github.com/shivam-sharma7/Horreum/blob/3cd396e31ac010de8facb1479d9042c127af700d/webapp/src/domain/schemas/AllSchema.tsx#L122
> 
> Same as above for : https://github.com/shivam-sharma7/Horreum/blob/3cd396e31ac010de8facb1479d9042c127af700d/webapp/src/domain/schemas/AllSchema.tsx#L129
> 
> Seeing that we need to know the total number of schemas for the pagination to work correctly, we also need to modify the REST API to return the total `count` of schemas

any example?

webapp/src/domain/schemas/AllSchema.tsx
shivam-sharma7: and we need to make change like- https://github.com/Hyperfoil/Horreum/blob/0c1147ff9d591dcfa655aeed09052e0694f1d502/horreum-model/src/main/java/io/hyperfoil/tools/horreum/api/ReportService.java#L74

johnaohara:
@shivam-sharma7 having chatted offline, I will modify the endpoint to return a `SchemaQueryResult` object

That will contain an `schemas:Array<Schema>` field and a `count` field

That should provide the necessary REST API endpoint to finish this implementation

shivam-sharma7:
@johnaohara see the latest changes.

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -108,6 +128,15 @@ export default function AllSchema() {
                      
                     />
                 </CardBody>
+                <CardFooter style={{ textAlign: "right" }}>
+                    <Pagination
+                        itemCount={schemas?.length || 0}
```
johnaohara, line 133:
```suggestion
                        itemCount={schemas?.count || 0}
```

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -23,6 +24,25 @@ export default function AllSchema() {
     document.title = "Schemas | Horreum"
     const dispatch = useDispatch<SchemaDispatch>()
 
+    const [page, setPage] = useState(1)
+    const [perPage, setPerPage] = useState(20)
+    const [direction, setDirection] = useState<SortDirection>("Ascending")
+    const pagination = useMemo(() => ({ page, perPage, direction }), [page, perPage, direction])
+    const [schemas, setSchemas] = useState<Schema[]>()
```
johnaohara, line 31:
```suggestion
    const [schemas, setSchemas] = useState<SchemaQueryResult>();
```

johnaohara (changes requested):
The Type for schemas has now changed, so the UI needs to reference the fields in this new Type

johnaohara:
@shivam-sharma7 is looking good, the failing tests just need fixing now. 

I can't comment on the PR directly as these lines have not been changed, but also need modifying

- `webapp/src/domain/schemas/actions.ts`;
  -  [line 39](https://github.com/shivam-sharma7/Horreum/blob/388/webapp/src/domain/schemas/actions.ts#L39) needs changing to reference the `Array<Schema>`  from the `SchemaQueryResult` response; e.g. `response => dispatch(loaded(response.schemas)),`
- `webapp/src/domain/schemas/AllSchema.tsx`;
  - [line 106 ](https://github.com/shivam-sharma7/Horreum/blob/388/webapp/src/domain/schemas/AllSchema.tsx#L106) needs removing
  - [line 119](https://github.com/shivam-sharma7/Horreum/blob/388/webapp/src/domain/schemas/AllSchema.tsx#L119) change to `schemas={schemas?.schemas || []}`
  - [line 126](https://github.com/shivam-sharma7/Horreum/blob/388/webapp/src/domain/schemas/AllSchema.tsx#L126) change to `data={schemas?.schemas || []}`

johnaohara:
@shivam-sharma7 looks like the linter is failing the UI build: https://github.com/Hyperfoil/Horreum/actions/runs/4892318596/jobs/8733938685?pr=510#step:6:427

johnaohara:
@shivam-sharma7 that shouldn't fail the build, they should just be warnings. I can't see why the build failed

johnaohara:
@shivam-sharma7 ah, the clue is in the previous line `Treating warnings as errors because process.env.CI = true` so, we need to fix the linter warnings: https://github.com/Hyperfoil/Horreum/actions/runs/4892318596/jobs/8733938685?pr=510#step:6:427

shivam-sharma7:
@jesperpedersen what do you think?

shivam-sharma7:
> @shivam-sharma7 ah, the clue is in the previous line `Treating warnings as errors because process.env.CI = true` so, we need to fix the linter warnings: https://github.com/Hyperfoil/Horreum/actions/runs/4892318596/jobs/8733938685?pr=510#step:6:427

Fixed ✅

johnaohara:
@shivam-sharma7  looks good, many thanks

johnaohara (approved):

webapp/src/domain/schemas/AllSchema.tsx
```diff
@@ -1,11 +1,11 @@
 import { useMemo, useEffect, useState } from "react"
 import { useSelector } from "react-redux"
 import { useDispatch } from "react-redux"
-import { Card, CardHeader, CardBody, PageSection } from "@patternfly/react-core"
+import { Card, CardHeader, CardFooter, CardBody, PageSection,Pagination } from "@patternfly/react-core"
```
jesperpedersen, line 4: Missing space

jesperpedersen (changes requested):

webapp/src/domain/schemas/AllSchema.tsx
shivam-sharma7: done

jesperpedersen:
Rebased, and merged.

PR #521, opened 2023-05-11
DTO refactor, Horreum needs to separate between it's internal data model and the external

stalep:
- All the classes in -api have been replaced
- The horreum-model was moved into -backend
- Entities are now named EntityNameDAO
- Added mapper functionality located in -backend


jesperpedersen:
I don't like EntityNameDAO at all - rename to EntityNameBean for now

stalep:
> I don't like EntityNameDAO at all - rename to EntityNameBean for now

Can you explain why? I had a look at several resources and DAOs seemed to be recommended over Bean.

jesperpedersen:
Its too close to DTO

stalep:
> Its too close to DTO

We have nothing named DTO so there should not be any confusion.

jesperpedersen:
Well, we can't call them EJB...

The default name for an Entity should be the name of the object itself - so EntityName - doing `@Table()` is just weird. Ideally the entire model would be locked in by the database and the application layer would have to adapt to that. That way DBAs to use database native functionality to optimize such as partitioning.

An Entity is def not a DTO - since that is a representation on a wire. And, since DAO is too close IMHO it is a no go.

jesperpedersen:
Rebased, and merge - due the Gang of Four.

PR #545, opened 2023-06-04
swagger-ui enable

shivam-sharma7:
## Changes proposed
image


jesperpedersen:
The diff is very big...

How do we turn it off for production environments ?

johnaohara:
the swagger-ui is enabled by default in quarkus dev mode when a dependency on `quarkus-smallrye-openapi` is added, and is not included in the runtime build, so is not available in PROD. 

We already depend on `quarkus-smallrye-openapi` so, this functionality is already available.. What has changed in the poms? I can not see the changes because the poms appear to have been re-formatted

PR #551, opened 2023-06-08
Algo change to render components

shivam-sharma7:
Algo change to render components


jesperpedersen:
Rebased, and merged.

PR #555, opened 2023-06-08
Remove explicit netty dependency

johnaohara:
We no longer need to import an explicit netty dependency, and should use the transitive dependency from Quarkus. Otherwise there might be unexpected behaviour


jesperpedersen:
Rebased, and merged.

PR #562, opened 2023-06-12
Implement Client upload and experiment test. Improve typesafety in API

johnaohara:
The PR introduces a client side Integration test that creates a new test, schema and experiment definition.  It then uploads run data, waits for datasets to be produced and then runs an Experiment. 

The API has been modified for this test so that it is type-safe.


jesperpedersen:
Rebased, and merged.

PR #576, opened 2023-06-29
Fix new test import for a Horreum instance that does not have test de…

johnaohara:


jesperpedersen:
Rebased, and merged.

PR #582, opened 2023-07-02
add node ci for front-end

shivam-sharma7:
add node ci for front-end

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -1,0 +1,24 @@
+name: Node.js GitHub Action
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: '18'
+
+    - name: Install dependencies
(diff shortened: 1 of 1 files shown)


johnaohara:
Hi @shivam-sharma7. Why do we need to re-test building the front-end?  
Compiling the front-end is already tested in the `main.yml` workflow.  The main.yaml workflow also tests generating the typescript client api and compiling the front-end with the auto-generated api, which this new workflow does not do

shivam-sharma7:
> Hi @shivam-sharma7. Why do we need to re-test building the front-end? Compiling the front-end is already tested in the `main.yml` workflow. The main.yaml workflow also tests generating the typescript client api and compiling the front-end with the auto-generated api, which this new workflow does not do

@johnaohara This workflows is for specialy for node dependencies like `npm install` and stuff which `main.yml` does't do this is basically for front-end dev who work with UI and stuff this is keep track

johnaohara:
@shivam-sharma7 `main.yaml` already covers this case, with a stricter command than `npm install`, it calls `npm ci` which will fail the build if the package versions do not match in `package-lock.json` and `package.json` : https://github.com/Hyperfoil/Horreum/actions/runs/5435864569/jobs/9885303737#step:6:360

`main.yaml` also builds the frontend: https://github.com/Hyperfoil/Horreum/actions/runs/5435864569/jobs/9885303737#step:6:383

We don't need to dupliate this check

PR #612, opened 2023-08-03
build infra before backend

stalep:
Backend depends on infra and without this change backend failed to build for me.


shivam-sharma7:
@stalep which OS are you using?

stalep:
Fedora

PR #618, opened 2023-08-11
common refractor

shivam-sharma7:

- ### [x] All new and existing tests passed.

### DIFF
--- a/webapp/src/components/EnumSelect.tsx
+++ b/webapp/src/components/EnumSelect.tsx
@@ -9,21 +9,21 @@ type EnumSelectProps = {
     isDisabled?: boolean
 }
 
-export default function EnumSelect(props: EnumSelectProps) {
+export default function EnumSelect({options, selected, onSelect, isDisabled}: EnumSelectProps) {
     const [isOpen, setOpen] = useState(false)
     return (
         <Select
             isOpen={isOpen}
             onToggle={setOpen}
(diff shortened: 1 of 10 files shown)


webapp/src/components/ActionMenu.tsx
```diff
@@ -55,7 +55,7 @@ export function useShareLink(config: ShareLinkConfig): MenuItem<ShareLinkConfig>
                         isOpen={shareLinkModalOpen}
                         onClose={() => setShareLinkModalOpen(false)}
                         isTester={isTester}
-                        link={config.token ? config.tokenToLink(props.id, config.token) : ""}
```
johnaohara, line 58: Why do we want to replace all ternary operator(s) with boolean operator(s)?  This fundamentally changes the evaluation logic and can introduce subtle bugs depending on the state of the 2nd operand, especially if that operand is coerced to "falsy"?

stalep:
> > Expanding the properties doesn't make sense to me. What is your reason to change this?
> > Imo this refactor makes it harder to read the code than before.
> 
> @stalep harder mean what @johnaohara suggest to me change and rest of the thing change by following current code style, specially `react with typescript`.

imo it's "easier" to read props.propetyName, as it tells me this comes from that property.

shivam-sharma7:
We don't need to use thing like `props.` we can access directly.

PR #620, opened 2023-08-14
removed json properties on DAO objects, had to annotate some with Jso…

stalep:
…nIgnoreType to prevent Jackonmapper to run wild


stalep:
To prevent Jackson to go wild (recursive) I had to add JsonIgnoreType on several of the DAOs.
Also, with this change the AlertingServiceTest.testChangeDetection:166->testSerialization:206 fails to deserialize the ChangeDAO.Event object as Jackson is no longer scanning it (this is what I assume happens).

So, either we change the MessageBus to use Change.Event objects instead or we need to wait with this refactor until we've refactored the MessageBus.

stalep:
The second commit fixes the test issues

jesperpedersen:
Move it out of draft once you are ready for a review

jesperpedersen:
Rebased, and merged.

PR #635, opened 2023-09-14
Added fk constraint from datapoint to dataset.

### LINKED ISSUE #591: Change Detection Variable Deletion causes a 500 error
**Describe the bug**
In a test that I'm editing on Red Hat's internal Horreum instance (https://horreum.corp.redhat.com/test/212), when deleting a variable in change detection, upon clicking "Save", a 500 error from the server occurs, and nothing is displayed to the user to indicate something went wrong.  No details are given in the 500 error that can be seen on the client browser.

**To Reproduce**
TBD

Simply making a new test, then adding a change detection variable, saving, then removing it does NOT reproduce this bug

whitingjr:
## Fixes Issue

fixes #591 

## Changes proposed

This PR adds a DML statement to clear out all orphaned datapoint rows. The DDL to create a foreign key constraint.

### DIFF
--- a/horreum-backend/src/main/resources/db/changeLog.xml
+++ b/horreum-backend/src/main/resources/db/changeLog.xml
@@ -4090,4 +4090,14 @@
             CREATE TRIGGER before_variable_delete BEFORE DELETE ON variable FOR EACH ROW EXECUTE FUNCTION before_variable_delete_func();
         </sql>
     </changeSet>
+
+    <changeSet id="107" author="jwhiting">
+        <sql>
+            delete from datapoint dp where dp.dataset_id not in (select ds.id from dataset ds);
+        </sql>
+        <addForeignKeyConstraint constraintName="fk_datapoint_dataset_id"
+                                 baseTableName="datapoint" baseColumnNames="dataset_id"
+                                 referencedTableName="dataset" referencedColumnNames="id" />
+    </changeSet>
+
 </databaseChangeLog>


horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -4090,4 +4090,14 @@
             CREATE TRIGGER before_variable_delete BEFORE DELETE ON variable FOR EACH ROW EXECUTE FUNCTION before_variable_delete_func();
         </sql>
     </changeSet>
+
+    <changeSet id="107" author="jwhiting">
+        <sql>
+            delete from datapoint dp where dp.dataset_id not in (select ds.id from dataset ds);
```
johnaohara, line 4096: i don't think we want to automate this, but handle it as a one-off manual process. It is specific to our env. adn this could break quietly future systems. @jesperpedersen thoughts?

horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -4090,4 +4090,14 @@
...
+        </sql>
+        <addForeignKeyConstraint constraintName="fk_datapoint_dataset_id"
+                                 baseTableName="datapoint" baseColumnNames="dataset_id"
+                                 referencedTableName="dataset" referencedColumnNames="id" />
+    </changeSet>
```
johnaohara, line 4101: we want to drop the `fk_dataset_run_id ` constraint as well

johnaohara (changes requested):

jesperpedersen:
This is an one-off - so no we don't want it in the master/stable branches.

We need migration scripts between versions if necessary.

jesperpedersen:
Rejected

johnaohara:
@jesperpedersen we still need the new FK constraint, and remove to the old constraint

jesperpedersen:
We can do those - either in the Java layer or as a rule

johnaohara:
@jesperpedersen this PR contained a rule to add the new FK constraint, it was just missing the removal. idk why you rejected the PR?

whitingjr:
#638 was created to continue this effort. This time with a commit that has changes in line with comments above.

PR #638, opened 2023-09-14
Added fk constraint from datapoint to dataset

### LINKED ISSUE #591: Change Detection Variable Deletion causes a 500 error

whitingjr:
## Fixes Issue

fixes #591 

## Changes proposed

This PR adds a DDL statement to remove the existing foreign key constraint named `fk_dataset_run_id`. Also adds the Liquibase configuration that will create a new constraint. A constraint from datapoint table to dataset.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java
@@ -1219,12 +1219,7 @@ public class AlertingServiceImpl implements AlertingService {
    @WithRoles(extras = Roles.HORREUM_SYSTEM)
    @Transactional
    void onDatasetDeleted(DataSet.Info info) {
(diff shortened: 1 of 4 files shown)


jesperpedersen:
Can you rebase ?

jesperpedersen:
Seeing

```
	Caused by: org.postgresql.util.PSQLException: ERROR: update or delete on table "dataset" violates foreign key constraint "fk_datapoint_dataset_id" on table "datapoint"
  Detail: Key is still referenced from table "datapoint".
```

during `mvn package`

jesperpedersen:
Same failures as CI

PR #663, opened 2023-09-19
fixes #653, added bean validation to the rest data objects

### LINKED ISSUE #653: Missing data Notification Rule name is not mandatory
The rule "Name" field for Missing data notifications is not mandatory

image

This is inconsistent with other forms

stalep:

### DIFF
--- a/horreum-api/pom.xml
+++ b/horreum-api/pom.xml
@@ -84,7 +84,7 @@
             <plugin>
                 <groupId>org.openapitools</groupId>
                 <artifactId>openapi-generator-maven-plugin</artifactId>
-                <version>6.0.0</version>
+                <version>7.0.1</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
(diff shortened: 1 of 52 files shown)


jesperpedersen:
@ia3andy Thanks Andy !

johnaohara:
The web UI failures are caused by a bug in `openapi-generator-maven-plugin`: have pushed an update https://github.com/Hyperfoil/Horreum/pull/663/commits/3e45d062ce32db739e7443ed1cb023f8d762115c

Also need to look at the workflow for creating a new test, as users can currently create a new test and add experiments etc to it before the test is saved to the db, and therefore in the client side cache.  So we either need to force a workflow where the test is saved before being configured, or we need to client side code to update the client side cache, so we can enforce a test against `ExperiementProfile` or `View`

jesperpedersen:
Squashed, rebased and merged.

Issue #664, opened 2023-09-19
Refactor out the ui services out of -api


stalep:
To make a clear distinction of what is included in the "client" and "ui" api we can split the ui specific services out to a -ui module that depends on -api.
This enables us to easily separate the generation of the openapi files.

johnaohara:
As mentioned on chat,

everything with JsonNode. ArrayNode, ObjectNode types are problematic, and we should make sure we try to split them out into a UI-api in the first instance where we can

we should also look at where we return `jakarta.ws.rs.core.Response` and probably change to `org.jboss.resteasy.reactive.RestResponse`, because we can have a typed response.
`jakarta.ws.rs.core.Response` is not generic

johnaohara:
And also we should validate the non-ui api openapi schema,  set this to `false`: 
https://github.com/Hyperfoil/Horreum/blob/master/horreum-api/pom.xml#L98

jesperpedersen:
Ideally we should only use "core" data types to Java or the platform.

It could be good to spend a bit of time to look at other platforms way of doing it such that we make sure that we can make clients work across programming languages

jesperpedersen:
I have looked at how we integrate with other programming languages and frameworks.

For all the JSON data I think we have to rely on `java.lang.String`, and then we need to validate in each of the API endpoints.

I would prefer the Jakarta classes, but the RestEasy one is ok as well

PR #674, opened 2023-09-25
Issue 664

stalep:


stalep:
Still needs tsx fixes, gives an idea of the -ui and -api split

jesperpedersen:
Can you squash ?

stalep:
I've tested import of tests and schemas that all work, I've also added views to tests.
The client openapi generated spec for Horreum is in the `horreum-api/generated/` folder.
There might be some ui/web issues that might pop up, but for what I've tested it's ok.

jesperpedersen:
This needs approval from @johnaohara as well

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Test.java
```diff
@@ -28,9 +28,6 @@ public class Test {
     @Schema(implementation = String[].class)
     public JsonNode fingerprintLabels;
     public String fingerprintFilter;
-    @NotNull
-    @JsonProperty(required = true)
-    public Collection<View> views;
```
jesperpedersen, line 33: Maybe we should take this separate

jesperpedersen:
I think the main discussion is about the `JsonNode <-> String` thing for this pull request.

We can take it on this pull request or start a discussion on it

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Test.java
stalep: View staying in -api caused a lot of changes to the tsx files. This actually made sense logically and also caused us to do less changes. win win :)

stalep:
> I think the main discussion is about the `JsonNode <-> String` thing for this pull request.
> 
> We can take it on this pull request or start a discussion on it

There are not many other options unfortunately. We can't avoid Json as a data structure but we can't use JsonNode if we want to have a openapi spec that's valid.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
```diff
@@ -71,6 +72,9 @@ public class ActionServiceImpl implements ActionService {
    @Inject
    EncryptionManager encryptionManager;
 
+   @Inject
+   TestServiceImpl testService;
```
johnaohara, line 76: Minor nitpick, This could be the api, which allows CDI to decide which impl to inject. IDK if we are ever going to have multiple impls, so prob not neccessary

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
```diff
@@ -219,6 +223,32 @@ public Action get(int id){
...
+         action.persist();
+      } else {
+         if (!action.active) {
+            ActionDAO.deleteById(action.id);
+            return null;
+         } else {
+            merge(action);
+         }
+      }
+      em.flush();
```
johnaohara, line 249: why do we need to flush? can't we let the transaction manager / ORM decide when to flush?

johnaohara:
> I think the main discussion is about the `JsonNode <-> String` thing for this pull request.
> 
> We can take it on this pull request or start a discussion on it

Issue #664 is around splitting the api into "data/analytics" and "api" modules.  The "data/analystics" module that other languages will consume definitely should not expose jackson datatypes, but the UI api can, as only the UI client code will interact with that api. 

Atm, changing some of the endpoints that expose/consume JsonNode, ArrayNode etc cascades a lot of changes in .tsx.  So as an interim, I think this is a good solution

webapp/src/domain/reports/TableReportConfigPage.tsx
```diff
@@ -272,7 +272,7 @@ export default function TableReportConfigPage() {
                                             owner: "",
                                             access: 0,
                                             notificationsEnabled: false,
-                                            views: [],
+                                            //views: [],
```
johnaohara, line 275: delete comment

webapp/src/domain/runs/DatasetComparison.tsx
```diff
@@ -23,6 +23,9 @@ import PrintButton from "../../components/PrintButton"
 import FragmentTabs, { FragmentTab } from "../../components/FragmentTabs"
 
 import { renderValue } from "./components"
+import * as actions from "../tests/actions";
```
johnaohara: can we expand the '*' to explicitly import functions

webapp/src/domain/tests/General.tsx
```diff
@@ -51,7 +51,7 @@ export default function General({ test, onTestIdChange, onModified, funcsRef }:
                 name,
                 folder,
                 description,
-                views: [], // automatically insert default view in backend
+                //views: [], // automatically insert default view in backend
```
johnaohara, line 54: delete comment

johnaohara (changes requested):

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Schema.java
```diff
@@ -6,18 +6,25 @@
 
 import java.util.Collection;
 
+@org.eclipse.microprofile.openapi.annotations.media.Schema(name = "Schema",
```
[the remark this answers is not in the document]
stalep: Not possible I think since the class is also named Schema

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
stalep: Unfortunately the getTestForUpdate method is not exposed on the api.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
stalep: The add method had an explicit flush so I added it for update as well, but no good argument except that :)

stalep:
The go client creation work with the latest changes, but python fails with several errors. Trying to investigate.

jesperpedersen:
Screenshot 2023-10-02 at 13-05-59 Roadrunner Test Horreum

stalep:
> Screenshot 2023-10-02 at 13-05-59 Roadrunner Test Horreum

Some context please?

jesperpedersen:
This PR w/ the example data loaded - then frontpage

stalep:
> This PR w/ the example data loaded - then frontpage

Do you have a stacktrace?

jesperpedersen:
And, navigate to http://localhost:8080/test/11

stalep:
I do not think those errors are caused by this pr

jesperpedersen:
This error isn't on `master`

jesperpedersen:
Rebased, squashed, and merged.

Issue #567, opened 2023-06-13
Blank Preview and Report  page


whitingjr:
**Describe the bug**

blank page where there should be a Report.

The Javascript console reports this error
```

TypeError: can't convert null to object
    children TableReportView.tsx:442
    children TableReportView.tsx:439
    children TableReportView.tsx:432
    td TableReportView.tsx:422
    React 7
    unstable_runWithPriority scheduler.production.min.js:18
    React 5
    rd TableReportPage.tsx:51
    rd TableReportPage.tsx:51
    Fl React
    unstable_runWithPriority scheduler.production.min.js:18
    React 4
    unstable_runWithPriority scheduler.production.min.js:18
    React 6
    361 index.tsx:9
    361 main.5dbe4aba.chunk.js:1
    f 4504761:1
    a 4504761:1
    t 4504761:1
    <anonymous> main.5dbe4aba.chunk.js:1
react-dom.production.min.js:216:199
    React 5
    unstable_runWithPriority scheduler.production.min.js:18
    React 4
    unstable_runWithPriority scheduler.production.min.js:18
    React 5
    rd TableReportPage.tsx:51
    rd TableReportPage.tsx:51
    Fl React
    unstable_runWithPriority scheduler.production.min.js:18
    React 4
    unstable_runWithPriority scheduler.production.min.js:18
    React 6
    361 index.tsx:9
    361 main.5dbe4aba.chunk.js:1
    f 4504761:1
    a 4504761:1
    t 4504761:1
    <anonymous> main.5dbe4aba.chunk.js:1
```

**To Reproduce**

Add a Component, set the function with a value that does not exist. 

**Version**

```
0.7.9
```

**Java**

```
11
```

**PostgreSQL**

```
psql --version
```

**OS**

Linux

**Browser**

v114

whitingjr:
this occurs with report id 4504761

whitingjr:
This issue occurs due to a Report Component attempting to work with a Label value that is null in a Dataset.
This happens when an old test result is used for creating the Report. The Dataset for that Run will not exist. Hence a null value for the Component in the Report.
 The missing value should be handled better. Allowing the page to display and be edited.

jesperpedersen:
You can open a pull request against 0.7.x also

PR #678, opened 2023-09-27
Added change to handle null value. 

whitingjr:
in the situation there is no dataset associated with a Component. Added testcase for this issue.

## Fixes Issue

fixes issue #567 

## Changes proposed

This change will add a null reference check to filter the Report references.  These exist in the situation when a report Component has an associated Label but for whatever reason there is no dataset. It can happen.


jesperpedersen:
Merged.

PR #747, opened 2023-10-13
remove javax dependency

shivam-sharma7:


jesperpedersen:
Did you verify the UI ?

`javax` is def not correct - but maybe it should be a `provided` `jakarta` dependency instead

shivam-sharma7:
@jesperpedersen Thanks for the hint, rebased and pushed

jesperpedersen:
Merged.

PR #753, opened 2023-10-13
Use specific `cmd` to install dependencies by quinoa

shivam-sharma7:


jesperpedersen:
What is this for ?

shivam-sharma7:
> What is this for ?

By default quinoa use `npm ci` which is install packages from .lock files and sometime this is give peer dep issue to avoid this we can use our own cmd like `npm install` what we use and this is download packages from `package.json file` and keep up to date with `.lock` file

shivam-sharma7:
don't mind but happy to see you on Sunday,

jesperpedersen:
Merged.

PR #746, opened 2023-10-13
Add HTTPS endpoint to keycloak test container

barreiro:

PR #759, opened 2023-10-17
Add HTTPS endpoint to keycloak test container with openssl certificate

barreiro:
This is what an alternative to #746 would look like when using openssl to pre-generate the certificate.


jesperpedersen:
Went with the BouncyCastle version

Issue #763, opened 2023-10-17
POSTing a schema to `/api/schema` creates new schema without transformers


johnaohara:
If a user exports a schema, and then calls POST to `/api/schema/`, the exported schema is imported but without any transformers defined.

The api should report an error when an exported schema is posted to the `/api/schema` endpoint, or persist the schema (and trasnformers) correctly

stalep:
The logic we have atm is very crude, I'm looking at it, but one situation is not clear cut. Lets say we import a schema; the id given does not exist (even though it's on the import), then I'm we should clear the id and try to import as a new schema. 
But, if we then see that the uri is already taken, should we merge with the matching uri schema or throw an exception?

johnaohara:
What about the scenario where someone exports from one instance `SchemaA.id = 101; SchemaA.urn = urn:my-benchmark:0.1` and imports into another instance where `CompletelyDifferentSchema.id = 101; CompletelyDifferentSchema.urn = urn:my-benchmark:0.1`? 

If we overwrite without confirmation from the user, there is the possibility of completely messing up all the existing data 

We don't currently have a way of rolling back changes: https://github.com/Hyperfoil/Horreum/issues/290

johnaohara:
So, We should probably have 2 endpoints, with clearly defined behaviour, i.e. 
- import schema -> fails if it detects same ID/URN
- update schema -> confirms that use wants to update, possibly showing existing vs import

stalep:
We could have a warning when the user updates with an import matching av existing uri it will override it?

jesperpedersen:
I think for schema it makes sense to have 2 endpoints for `new` and `update` (comment copied from #772)

PR #772, opened 2023-10-19
improve logic when importing schemas

stalep:

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
@@ -147,17 +147,7 @@ public class SchemaServiceImpl implements SchemaService {
    @Transactional
    @Override
    public Integer add(Schema schemaDTO){
-      if (schemaDTO.uri == null || Arrays.stream(ALL_URNS).noneMatch(scheme -> schemaDTO.uri.startsWith(scheme + ":"))) {
-         throw ServiceException.badRequest("Please use URI starting with one of these schemes: " + Arrays.toString(ALL_URNS));
-      }
-      SchemaDAO byName = SchemaDAO.find("name", schemaDTO.name).firstResult();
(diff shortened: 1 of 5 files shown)


stalep:
This fixes #763 but the question is if we should change the api while we're at it?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
```diff
@@ -190,6 +180,20 @@ public Integer add(Schema schemaDTO){
       return schema.id;
    }
 
+   private void verifyNewSchema(Schema schemaDTO) {
+      if (schemaDTO.uri == null || Arrays.stream(ALL_URNS).noneMatch(scheme -> schemaDTO.uri.startsWith(scheme + ":"))) {
+         throw ServiceException.badRequest("Please use URI starting with one of these schemes: " + Arrays.toString(ALL_URNS));
+      }
+      SchemaDAO byName = SchemaDAO.find("name", schemaDTO.name).firstResult();
+      if (byName != null && !Objects.equals(byName.id, schemaDTO.id)) {
+         throw ServiceException.serverError("Name already used");
```
johnaohara, line 189: This should probably be changed to a BadRequest, rather than serverError

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
```diff
@@ -190,6 +180,20 @@ public Integer add(Schema schemaDTO){
...
+      }
+      SchemaDAO byUri = SchemaDAO.find("uri", schemaDTO.uri).firstResult();
+      if (byUri != null && !Objects.equals(byUri.id, schemaDTO.id)) {
+         throw ServiceException.serverError("URI already used");
```
johnaohara, line 193: same as above

jesperpedersen:
I think for schema it makes sense to have 2 endpoints for `new` and `update`

jesperpedersen:
Conflict resolved, rebased, and merged.

PR #792, opened 2023-10-22
Update  contributor section

shivam-sharma7:
image


shivam-sharma7:
I know we can have more, but we need to add here only very active contributor.

shivam-sharma7:
I know we can have more, but we need to add here only very active contributor.

jesperpedersen:
We value all contributions - no matter when they were made.

Rejected.

PR #837, opened 2023-10-29
Migrate deprecated javax.security.jacc packages to jakarta.security.jacc

shivam-sharma7:


jesperpedersen:
Decided to remove native instead

PR #847, opened 2023-10-30
fixes #800

### LINKED ISSUE #800: `SchemaService.dropToken()` should not return a String value and is incorrect HTTP verb
`SchemaService.dropToken()` should be a `@DELETE` operation and should not return a value

stalep:
All tests passes locally, do not see the need for explicit tests for this change.


jesperpedersen:
Merged.

PR #848, opened 2023-10-30
Fix toEpochMillis for JS Dates: Fixes #840

### LINKED ISSUE #840: Executed timestamp difference incorrect
The `Executed` column for Runs and Datasets is not being calculated correctly.

Load example data, and navigate to the Runs viw. The start timestamp is `2022-11-25 15:08:51`, but the `Executed` column is displaying `54 years ago`

Screenshot from 2023-10-30 07-12-33

Screenshot from 2023-10-30 07-12-25

johnaohara:


jesperpedersen:
Merged.

PR #850, opened 2023-10-30
Display null for json path expression evaluations: Fixes #843

### LINKED ISSUE #843: Try It! feature of Schema Label Extractor form does not execute against run data
Tht "Try It!" feature of the schema -> labels -> executor displays the following modal form:

image

Clicking on "Execute" does not show any results, it is not clear if the path expression is executed, or the call fails

johnaohara:


jesperpedersen:
Merged.

PR #863, opened 2023-11-02
Allow backport of #854 to 0.9.x branch

johnaohara:
Allow backport of #854 into 0.9.x branch by re-ordering of DB changesets.  There has not been a release of 0.10 yet, so unless someone is running a version of Horreum built from main branch, this should be "safe"

Issue #854, opened 2023-10-31
Duplicate primary key generated for PeristentLog, fails processing


johnaohara:
The Id generation strategy for `PersistentLogDAO` is defaulting to `AUTO`, which is generating duplicate ID's in some PersistentLog implementations. 

The PK constraint failure is causing a TX rollbackfor some data processing operations, and therefore some operations are failing to process data, e.g. recalculation of datapoints for change detection

Server side error log:

```
2023-10-31 12:37:01,939 ERROR [io.hyp.too.hor.bus.TaskQueue] (executor-thread-7935) Error executing task in the queue for test 254: org.hibernate.exception.ConstraintViolationException: could not execute statement [ERROR: duplicate key v
alue violates unique constraint "calculationlog_pkey"] [insert into DatasetLog (dataset_id,level,message,source,testid,timestamp,id) values (?,?,?,?,?,?,?)]
        at org.hibernate.exception.internal.SQLStateConversionDelegate.convert(SQLStateConversionDelegate.java:95)
        at org.hibernate.exception.internal.StandardSQLExceptionConverter.convert(StandardSQLExceptionConverter.java:56)
        at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:108)
        at org.hibernate.engine.jdbc.internal.ResultSetReturnImpl.executeUpdate(ResultSetReturnImpl.java:278)
        at org.hibernate.engine.jdbc.mutation.internal.AbstractMutationExecutor.performNonBatchedMutation(AbstractMutationExecutor.java:107)
        at org.hibernate.engine.jdbc.mutation.internal.MutationExecutorSingleNonBatched.performNonBatchedOperations(MutationExecutorSingleNonBatched.java:40)
        at org.hibernate.engine.jdbc.mutation.internal.AbstractMutationExecutor.execute(AbstractMutationExecutor.java:52)
        at org.hibernate.persister.entity.mutation.InsertCoordinator.doStaticInserts(InsertCoordinator.java:169)
        at org.hibernate.persister.entity.mutation.InsertCoordinator.coordinateInsert(InsertCoordinator.java:111)
        at org.hibernate.persister.entity.AbstractEntityPersister.insert(AbstractEntityPersister.java:2784)
        at org.hibernate.action.internal.EntityInsertAction.execute(EntityInsertAction.java:102)
        at org.hibernate.engine.spi.ActionQueue.executeActions(ActionQueue.java:635)
[...]

PR #864, opened 2023-11-02
Define sequences and ID generators for each PeristentLog subclasses: …

johnaohara:
This is the backport of #854 into 0.9.x branch.   This should only be merged if the reordering strategy in https://github.com/Hyperfoil/Horreum/pull/863 is acceptable, and should only be merged after #863 has been merged


jesperpedersen:
Merged.

PR #869, opened 2023-11-03
Fix mapping issue in change detection variable definitions: fixes #822

### LINKED ISSUE #822: change detection calculation functions not working and not saving
**Describe the bug**

In change detection I'm setting a variable using two labels, both of which are floating points. I'm trying to simply multiply them to understand how to write the function, but everything I try ends up with it telling me that it "Cannot turn {"Kernel Average ms":258.12,"Kernel Standard Deviation %":7.973} into a floating-point value for variable Testing/TEST Kernel Average ms". I just have the variable defined with those two labels: "Kernel Average ms" and "Kernel Standard Deviation %", and for testing I have tried the calculation:

`(input)=>input["Kernel Average ms"] * input["Kernel Standard Deviation %"]`

image

Running the recalculation with this function results in the error noted above reported in the logs. Then in the changes view, the chart for this variable is blank. Going back to the edit view, the calculation function I had added is completely gone.

image
image

I tried this again with a much simpler test, having only one floating point label in the variable and setting the calculation function to simply `value => value * 2`, and in this case I don't get any error logs on recalculating, but the change graph does not reflect values with the multiplier (only the base value of the label), and when I go back to edit the variable again for the change detection, the calculation function is always gone.

image

**Version**
Horreum (hosted) 0.9.4

**Browser**
Brave v1.59.117
(FWIW I did try with Brave "shields down" and had the same results)

johnaohara:
Backport for #834

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/mapper/VariableMapper.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/mapper/VariableMapper.java
@@ -11,7 +11,7 @@ public class VariableMapper {
 
     public static Variable from(VariableDAO variable) {
         return new Variable(variable.id, variable.testId, variable.name, variable.group,
-                variable.order, variable.labels,
+                variable.order, variable.labels, variable.calculation,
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

PR #879, opened 2023-11-09
Use `@quarkus 3.5.1`

shivam-sharma7:


shivam-sharma7:
CI fail not related, it's complaining for value which is not in use like `Fingerprints`

shivam-sharma7:
now Ci should be green, I rebase the PR

jesperpedersen:
Merged.

PR #912, opened 2023-11-15
correct config for keycloak admin client in dev mode : Fixes #911

### LINKED ISSUE #911: Keycloak admin client authenticating against wrong realm in dev mode
The configuration for the keycloak admin client is authenticating against the wrong realm, with incorrect credentials in dev mode

johnaohara:
Fixes #911


jesperpedersen:
Back-ported

Issue #779, opened 2023-10-19
Remove hashing WithRoles - configure TLS encryption between application and database


johnaohara:
Replace custom encryption for roles with TLS encryption between application and database

We need to;

- [ ] remove hashing code in application
- [ ] remove hashing code in DB `has_role` procedure
- [ ] remove `horreum.db.secret` property
- [ ] remove `dbsecret` database table
- [ ] allow enable ssl in dev/test mode
- [ ] update documentation on how to define a secure connection between Horreum and production database

johnaohara:
For dev mode we could do something similar to for the database connection: https://github.com/Hyperfoil/Horreum/pull/746

jesperpedersen:
We currently depend on [pgacrypto](https://www.postgresql.org/docs/current/pgcrypto.html) for this

PR #937, opened 2023-11-24
Remove db encryption with pgcrypto

barreiro:
Fixes: #779


johnaohara:
@barreiro has there been any further progress on this? thanks

johnaohara:
@barreiro the integration tests run with a built jar and therefore runs in prod mode by default.  We either need to enable ssl in the postgres container or do not enable by default for prod in `application.properties` and make sure it is highlighted in to docs that it is recommended to enable TLS in production.

I do not think creating and registering a self-signed cert for integration test will be simple

johnaohara:
@barreiro please can you squash commits

johnaohara:
LGTM, please squash and I will merge, thanks!

barreiro:
rebased and squashed

PR #994, opened 2023-12-20
NPE: Retransform Datasets

shivam-sharma7:

### DIFF
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/ValidationError.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/ValidationError.java
@@ -8,7 +8,7 @@ import java.util.Map;
 
 public class ValidationError {
     @Schema(description = "Schema ID that Validation Error relates to", example = "101")
-    public int schemaId;
+    public Integer schemaId;
     @JsonProperty(required = true)
     @Schema(description = "Validation Error Details")
     public ErrorDetails error;


johnaohara:
@shivam-sharma7 thanks for looking at this.  

It is valid to return `null` for the schema ID, where the schema is `null`. Setting the ID to an arbitrary value will likely introduce subtle bugs that we will need to guard against, e.g.  ` if ( schemaID != 0 ) { //do something}`

The root cause of this error is at this line: https://github.com/Hyperfoil/Horreum/blob/f2360cb4f64affa24d079cd88ee8b8e92ab9e84b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/mapper/ValidationErrorMapper.java#L16

`ve.getSchemaId()` returns null, which we try to assign to `dto.schemaId`, however `dto.schemaId` is a primitive type `int`: https://github.com/Hyperfoil/Horreum/blob/f2360cb4f64affa24d079cd88ee8b8e92ab9e84b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/ValidationError.java#L9

an `int` can not be null, The error message is being thrown by the JVM autoboxing.

To fix this, we need to change `dto.schemaId` from type `int` to `Integer`, which can be null

johnaohara (changes requested):
Please see my comment about types

johnaohara:
Merged, 

Thanks a lot for your contribution

Issue #785, opened 2023-10-19
Support Elasticsearch as a datasource


johnaohara:
- [x] Integrate an elastic back-end to connect to and query a remote elasticsearch instance
- [x] test configuration for different backends
- [x] UI configuration to allows users to provide required information to connect to elastic instance
- [x] Ensure isolation between tenants
- [x] API changes to allow passing in of index/id information to query documents in elastic

johnaohara:
Fixed in : https://github.com/Hyperfoil/Horreum/pull/1003

PR #1003, opened 2023-12-21
Introduce elascticsearch backend with querying and analysis

johnaohara:
Implements #785


horreum-web/package.json
```diff
@@ -46,7 +46,7 @@
     "uuid": "9.0.0"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "PORT=0 react-scripts start",
```
shivam-sharma7, line 49:
```suggestion
    "start": "PORT=0 react-scripts start",
```

Why? PORT=0

johnaohara: PORT=0 should bind the server to the first ephemeral port, this helps when you have more than one dev instance running

shivam-sharma7: Nice 👍

stalep (approved):

PR #1147, opened 2024-01-23
Detect when RLS has been a cause for error.

### LINKED ISSUE #406: Create RLS exception mapper to provide meaningful error messages to User
At present, RLS exceptions are bubbled back through the stack and the error ID is displayed in the UI.  The stack trace is logged but a Error ID is reported to the user. There is no information provided to the User about how to correct the issue, e.g. missing perms when trying to perform a certain action.

It would improve usability if we could intercept the exceptions from the database and provide a more meaningful error message to the user about how to correct what caused the error.

e.g., instead of reporting ;

```
Warning alert:Test access update failed

{
  "details": "Error id 48d3aa5f-dd62-46f4-a745-bfacebf4ef24-73",
  "stack": ""
}
```

Reporting the exception as ;

```
Warning alert: You are not defined as a `tester` for the `perf-team`. Please update your permissions and try again
```

### LINKED ISSUE #1145: Attempting update access on Run attemps using Test update access..
**Describe the bug**

The Run typescript object is currently calling [updateAccess](https://github.com/Hyperfoil/Horreum/blob/master/horreum-web/src/domain/runs/Run.tsx#L99) on line 99. Then invoking [apis.updateAccess](https://github.com/Hyperfoil/Horreum/blob/master/horreum-web/src/api.tsx#L259) which then fails if a Test with the same id is not present. Otherwise it updates the Test.

 Run.txs should be calling [apis.updateupdateRunupdateRunAccess](https://github.com/Hyperfoil/Horreum/blob/master/horreum-web/src/api.tsx#L354-L356)

How to recreate

Choose a Run (any Run) that does not have a Test with the same corresponding id. Then update the access property of the Run.
The following message is display.

`Failed to update test access, check your account has the necessary 'tester' role for the team.`

**Version**

```
0.10.7
```

**Java**

```
17
```

whitingjr:
Fixed calling the wrong apis function when updating Run access.

## Fixes Issue

fixes #406 
fixes #1145

## Changes proposed

- [x] Add catch block for unauthorized exception type.
- [x] Swich call to apis to execute the Run access update function.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
@@ -17,6 +17,7 @@ import io.hyperfoil.tools.horreum.server.WithToken;
 import io.quarkus.hibernate.orm.panache.PanacheQuery;
 import io.quarkus.panache.common.Page;
 import io.quarkus.panache.common.Sort;
+import io.quarkus.security.UnauthorizedException;
 import io.quarkus.security.identity.SecurityIdentity;
 
 import jakarta.annotation.security.PermitAll;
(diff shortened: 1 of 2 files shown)


johnaohara (approved):
LGTM

johnaohara:
Merged

Thanks for your contribution!

Issue #714, opened 2023-10-04
Migration Patternfly 4 to 5


shivam-sharma7:
**Feature idea.**
Horreum is using patternfly 4 and we need patternfly 5

Learn- https://www.patternfly.org/design-foundations/about

Steps-https://www.patternfly.org/get-started/upgrade/

willr3:
I started work on this so we can then implement #903

shivam-sharma7:
@willr3 You can take reference from here-https://github.com/infinispan/infinispan-console/commit/0fbd87e197c194fd9e5fdfe638f5a44b2f86f526

willr3:
I have a branch building with patternfly5 but we without UI tests I'm not certain it is working correctly
https://github.com/willr3/Horreum/tree/patternfly5

PR #1162, opened 2024-01-24
Patternfly5

willr3:
## Fixes Issue
Migrates to patternfly5 using the `deprecated` patternfly package where needed. We will still need to migrate off the deprecated package.

Fixes: #714

### DIFF
--- a/horreum-web/package.json
+++ b/horreum-web/package.json
@@ -5,9 +5,9 @@
   "type": "module",
   "dependencies": {
     "@monaco-editor/react": "4.6.0",
-    "@patternfly/patternfly": "4.224.4",
-    "@patternfly/react-core": "4.277.0",
-    "@patternfly/react-table": "4.113.5",
+    "@patternfly/patternfly": "5.1.0",
+    "@patternfly/react-core": "5.1.1",
+    "@patternfly/react-table": "5.1.1",
     "@types/history": "4.7.11",
     "@types/jsonpath": "0.2.4",
     "@types/luxon": "3.2.0",
--- a/horreum-web/package-lock.json
+++ b/horreum-web/package-lock.json
@@ -9,9 +9,9 @@
       "version": "0.1.0",
       "dependencies": {
         "@monaco-editor/react": "4.6.0",
-        "@patternfly/patternfly": "4.224.4",
-        "@patternfly/react-core": "4.277.0",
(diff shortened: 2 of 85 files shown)


horreum-web/package.json
```diff
@@ -4,9 +4,9 @@
   "private": true,
   "dependencies": {
     "@monaco-editor/react": "4.6.0",
-    "@patternfly/patternfly": "4.224.4",
-    "@patternfly/react-core": "4.277.0",
-    "@patternfly/react-table": "4.113.5",
+    "@patternfly/patternfly": "^5.1.0",
+    "@patternfly/react-core": "^5.1.1",
+    "@patternfly/react-table": "^5.1.1",
```
shivam-sharma7, line 9:
```suggestion
    "@patternfly/patternfly": "5.1.0",
    "@patternfly/react-core": "5.1.1",
    "@patternfly/react-table": "5.1.1",
```
please, use specific versions

shivam-sharma7 (changes requested):

horreum-web/package.json
willr3: This is for a UI library, is there an issue with putting the version range in package.json and relying on package-lock to preserve the specific version?

willr3: I made the change and see `"vite-tsconfig-paths": "^4.2.3"` in `master` is that also being updated to follow a specific version?
https://github.com/Hyperfoil/Horreum/blob/master/horreum-web/package.json#L49

johnaohara: it should be yes

johnaohara: > This is for a UI library, is there an issue with putting the version range in package.json and relying on package-lock to preserve the specific version?

Until we have unit tests for the UI, this limits the probability of in compatible transient deps breaking things. It isn't ideal, but neither is having a lack of UI testing. We can revisit this once we have UI tests that will detect breaking changes

shivam-sharma7: @willr3 make sure your lock file is up to date with your current change

johnaohara:
I was going to backport this, but given the risk of introducing problems, I am going to hold off until the 0.12 release to give more time to test

johnaohara:
@willr3 this looks good now, thanks for fixing the issues.

Please can you squash and rebase, thanks

johnaohara:
Thanks a lot for your contribution!

PR #1183, opened 2024-01-26
Add endpoint to fetch the roles of an authenticated user: Fixes #1187

### LINKED ISSUE #1187: Provide Roles REST endpoint to retrieve user roles from the application
User roles are currently managed in Keycloak.  In order to manage the teams/roles within the application (https://github.com/Hyperfoil/Horreum/issues/780), the UI needs to be able to retrieve the users roles from the Horreum REST API instead of making a call to keycloal

barreiro:
Fixes: #1187

### DIFF
--- a/horreum-web/src/Login.tsx
+++ b/horreum-web/src/Login.tsx
@@ -4,7 +4,7 @@ import {AppContextType} from "./context/@types/appContextTypes";
 import {Button, Form, FormGroup, Modal, Spinner, TextInput} from "@patternfly/react-core";
 import {userApi} from "./api";
 import store from "./store";
-import {BASIC_AUTH, UPDATE_ROLES, AFTER_LOGOUT} from "./auth";
+import {BASIC_AUTH, UPDATE_ROLES, AFTER_LOGOUT, STORE_PROFILE, UPDATE_DEFAULT_TEAM} from "./auth";
 
 type LoginModalProps = {
     username: string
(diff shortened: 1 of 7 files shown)


horreum-web/src/Login.tsx
```diff
@@ -38,12 +38,18 @@ export default function LoginModal(props: LoginModalProps) {
...
+                        .then(roles => {
+                                alerting.dispatchInfo("LOGIN", "Log in successful", "Successful log in of user " + username, 3000)
+                                store.dispatch({type: UPDATE_ROLES, authenticated: true, roles: roles});
+                                userApi.searchUsers(username!).then(
+                                    userData => store.dispatch({type: STORE_PROFILE, profile: userData.filter(u => u.username == username).at(0)}),
+                                    error => alerting.dispatchInfo("LOGIN", "Unable to get user profile", error, 30000))
+                                userApi.defaultTeam().then(
+                                    response => store.dispatch({ type: UPDATE_DEFAULT_TEAM, team: response }),
+                                    error => alerting.dispatchInfo("LOGIN", "Cannot retrieve default team", error, 30000)
+                                )
```
johnaohara, line 52: What is the definition of a "successful" login?  we need all the user info for the UI to function properly, so if the call to `userApi.defaultTeam()` fails, I don't think we should display a "Log in successful" message.

At the moment the user could see;
"Successful log in of user " msg
followed by
"Cannot retrieve default team" msg

I think this would be confusing.

I think we need to chain the api calls in .then()s and only display "Login Successful" if all the calls do not error

horreum-web/src/Login.tsx
```diff
@@ -38,12 +38,18 @@ export default function LoginModal(props: LoginModalProps) {
...
-                                alerting.dispatchInfo("LOGIN", "Log in successful", "Successful log in of user " + userdata[0].username, 3000)
-                                store.dispatch({type: UPDATE_ROLES, authenticated: true, roles: [/* TODO: the roles fetched */]});
+                    
+                    userApi.getRoles()
+                        .then(roles => {
+                                alerting.dispatchInfo("LOGIN", "Log in successful", "Successful log in of user " + username, 3000)
+                                store.dispatch({type: UPDATE_ROLES, authenticated: true, roles: roles});
+                                userApi.searchUsers(username!).then(
+                                    userData => store.dispatch({type: STORE_PROFILE, profile: userData.filter(u => u.username == username).at(0)}),
+                                    error => alerting.dispatchInfo("LOGIN", "Unable to get user profile", error, 30000))
```
johnaohara, line 48: Could we add to the error msg to tell the user what to do in this situation? something like ". please contact your system administrator"

horreum-web/src/Login.tsx
```diff
@@ -38,12 +38,18 @@ export default function LoginModal(props: LoginModalProps) {
...
+                                userApi.defaultTeam().then(
+                                    response => store.dispatch({ type: UPDATE_DEFAULT_TEAM, team: response }),
+                                    error => alerting.dispatchInfo("LOGIN", "Cannot retrieve default team", error, 30000)
```
johnaohara, line 51: same comment as above, can we provide an action that the user can do to rectify this issue

johnaohara (changes requested):
Please see inline comments. Thanks

horreum-web/src/Login.tsx
barreiro: good question. a successful login would be to get `200 OK` response from the `roles` endpoint. the user profile and default team are optional in the frontend.
 
most likely, we'll have to revisit the login procedure in the near future.

johnaohara:
@barreiro please can you update the error message, then we can get this merged, thanks

barreiro:
@johnaohara I believe we can leave it as is, at least for now. The login requires a successful call to the `roles` endpoint, while if the other requests fail the UI is able to deal with it and the missing information is not shown. The user does not need to take any action.

johnaohara:
> @johnaohara I believe we can leave it as is, at least for now. The login requires a successful call to the `roles` endpoint, while if the other requests fail the UI is able to deal with it and the missing information is not shown. The user does not need to take any action.

If the user does not need to/can not take any action, then we shouldn't display an error message, it is confusing and users can not do anything about it.

johnaohara (approved):

PR #1212, opened 2024-01-29
 Added TestExport to avoid creating unstructured json, fixes #860

### LINKED ISSUE #860: TestService.importTest will not work if importing as a new Test, export should also be revisited
The Json structure does not support importing the data if it's importing to a different Horreum instance.
We should look at creating an object structure for both importing and exporting.
This will affect the API so ideally we should try to get it into 0.10...

stalep:
Importing new test should work now correctly. Currently Experiments&/Comparisons are not correctly imported.

fixes #860


horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/TestService.java
```diff
@@ -228,18 +229,18 @@ List<ExportedLabelValues> listLabelValues(@PathParam("id") int testId,
 
    @GET
    @Path("{id}/export")
-   @APIResponseSchema(value = String.class,
+   @APIResponseSchema(value = TestExport.class,
            responseDescription = "A Test defintion formatted as json",
            responseCode = "200")
-   String export(@PathParam("id") int testId);
+   TestExport export(@PathParam("id") int testId);
```
johnaohara, line 235: I think we should still keep this generic, but use OjectNode instead of String, so that we are not tied to a single TestExport structure and can possibly support multiple versions of TestExport.
I think we need to consider backwards compatibility between Horreum Versions

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/TestService.java
```diff
@@ -228,18 +229,18 @@ List<ExportedLabelValues> listLabelValues(@PathParam("id") int testId,
...
 
    @POST
    @Path("import")
    @APIResponse(responseCode = "204", description = "Import a new test")
    @RequestBody(content = @Content( mediaType = MediaType.APPLICATION_JSON,
-           schema = @Schema( type = SchemaType.STRING, implementation = String.class)) )
+           schema = @Schema( type = SchemaType.STRING, implementation = TestExport.class)) )
    @Operation(description="Import a previously exported Test")
-   void importTest( String testConfig);
+   void importTest(TestExport testExport);
```
johnaohara, line 243: Same comment about backwards compatability

johnaohara (changes requested):
Some comments;

1) Please can we have some user docs for test export/import?

2) One thought, how do we handle different versions of test import/export? Or how would we in the future?  One significant impact of this change is that we are bound to one version of an exported test for the version of Horreum running, so say you could not export a test from Horreum 0.11.1 and import into 0.13.4 if the test data structure changes between releases.

The old API would allow users to upload the json payload an Horreum had the ability to perform some form of check, but with this change we would return a HTTP 400 bad request if the JSON did not deserialise to a valid TestExport.

I think we should consider backwards compatibility between Horreum Versions. I am not saying we should make import/export backwards compatible now, but leave the possibility of supporting it in the future 


3) We need to consider datastores?  The default datastore is fine, but what about a Test with an Elastic backend etc?   We either need to throw an error, or tell the user to configure a datastore first, or serialize datastore with the test.  Serializing the datastore would require careful consideration about how to handle API tokens

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
```diff
@@ -307,57 +308,42 @@ public void onNewExperimentResult(ExperimentService.ExperimentResult result) {
...
-         for (JsonNode node : actions) {
-            if (!node.isObject()) {
-               throw ServiceException.badRequest("Test actions must be an array of objects");
+   void importTest(TestExport test) {
+      for (Action a : test.actions) {
+         try {
+            ActionDAO action = ActionMapper.to(a);
+            if (a.secrets != null) {
+               action.secrets = Util.OBJECT_MAPPER.readTree(encryptionManager.decrypt(a.secrets.textValue()));
+            } else {
```
johnaohara, line 333: encryptionManager uses `horreum.db.secret` as the password for encryption. if this is different between Horreum instances, the secrets encryption/decryption will not work

johnaohara: So, we need to think about/if we should serialize secrets, and provide a mechanism for redefining secrets in the new test

johnaohara:
> Currently Experiments&/Comparisons are not correctly imported.

Please can you open an issue to track this

stalep:
> Some comments;
> 
>     1. Please can we have some user docs for test export/import?
Yes!
> 
>     2. One thought, how do we handle different versions of test import/export? Or how would we in the future?  One significant impact of this change is that we are bound to one version of an exported test for the version of Horreum running, so say you could not export a test from Horreum 0.11.1 and import into 0.13.4 if the test data structure changes between releases.
> 
> 
> The old API would allow users to upload the json payload an Horreum had the ability to perform some form of check, but with this change we would return a HTTP 400 bad request if the JSON did not deserialise to a valid TestExport.

In reality there is no difference, before this change if someone would send a json that did not deserialize to a Test + the additional fields it would fail. With this change the JSON is a bit different, but it is typed and we avoid custom (de)serialization as it's handled by Quarkus.

> 
> I think we should consider backwards compatibility between Horreum Versions. I am not saying we should make import/export backwards compatible now, but leave the possibility of supporting it in the future

With using TestExport we can add additional fields later on and it would still work with older version.

> 
>     3. We need to consider datastores?  The default datastore is fine, but what about a Test with an Elastic backend etc?   We either need to throw an error, or tell the user to configure a datastore first, or serialize datastore with the test.  Serializing the datastore would require careful consideration about how to handle API tokens

I have not considered datastores at all tbh, we could also add datastores to TestExport...?

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/TestService.java
stalep: Personally I like typed APIs. Using ObjectNode would be an "articial" abstraction as it would be deserialized to TestExport anyway. If we add a field to TestExport in a later version an older version would still be accepted by Horreum, but it would not populate the new field.

johnaohara: The TestExport class can be mutated but adding, removing fields, changing field types and changing relationships in Entity graph that TestExport represents. deserialization breaks when you change types and relationships, and we would have not control over it

johnaohara: The api can still be typed, e.g.;

```
    @NotNull
    @JsonProperty( required = true )
    @Schema(type = SchemaType.OBJECT,
            oneOf = {
                TestExport_v1.class,
                    TestExport_v2.class
    })
    public ObjectNode export(@PathParam("id") int testId);
 ```

Comment (johnaohara) [reply horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/TestService.java:235]: See: https://horreum.hyperfoil.io/openapi/#tag/Config/operation/newDatastore as an example

**Comment by johnaohara** [comment]
> In reality there is no difference, before this change if someone would send a json that did not deserialize to a Test + the additional fields it would fail. With this change the JSON is a bit different, but it is typed and we avoid custom (de)serialization as it's handled by Quarkus.

Under the hood Quarkus is using ObjectMapper to serialize/deserialize to/from json, we can still do the same without the customer deserialization.  We can still enforce the type on the API, accepting different versions of concrete types.

If we don't we are locking an instance of Horreum to only accept Tests exported from another Horreum instance that has exported the same TestExport entity.

Adding fields might be fine, but if we removed required fields, change their type, change the relationship between test and other entities then it will break.

At the moment, we are effectively writing the DB structure to JSON. I don't see that staying stable between Horreum versions, and a change to the API to somthing like this, would still provide type safety, but give use the opportunity to evolve the entity graph contained in TestExport, whilst also supporting backwards compatibility;

```
    @NotNull
    @JsonProperty( required = true )
    @Schema(type = SchemaType.OBJECT,
            oneOf = {
                TestExport_v1.class,
                    TestExport_v2.class
    })
    public ObjectNode export(@PathParam("id") int testId);
```

**Comment by johnaohara** [comment]
> I have not considered datastores at all tbh, we could also add datastores to TestExport...?

We could, but how do we handle the secrets? This relates to secrets in the existing export as well.  We can't encrypt/decrypt between Horreum instances that have different `horreum.db.secret` properties.

We might not want to handle that problem in this PR, so we could just add Datastore to the output and open another issue to handle secrets in the exported TestExport definition

**Comment by stalep** [comment]
> If we don't we are locking an instance of Horreum to only accept Tests exported from another Horreum instance that has exported the same TestExport entity.
I think that's a fair limitation actually :)
> 
> Adding fields might be fine, but if we remove required fields, change their type, or change the relationship between the test and other entities then it will break.
Yes, and as I answered above it's a fair limitation imo, but I see your point.
> 
> At the moment, we are effectively writing the DB structure to JSON. I don't see that staying stable between Horreum versions, and a change to the API to something like this, would still provide type safety, but give us the opportunity to evolve the entity graph contained in TestExport, whilst also supporting backward compatibility;

Yes, I also see that changing in the not too distant future. What we then will end up with is a lot of parsing logic to support "old" versions of Horreum none might use...
I need to think about this..

> 
> ```
>     @NotNull
>     @JsonProperty( required = true )
>     @Schema(type = SchemaType.OBJECT,
>             oneOf = {
>                 TestExport_v1.class,
>                     TestExport_v2.class
>     })
>     public ObjectNode export(@PathParam("id") int testId);
> ```

**Comment by stalep** [comment]
> We might not want to handle that problem in this PR, so we could just add Datastore to the output and open another issue to handle secrets in the exported TestExport definition

Yeah, that's a good suggestion.

**Comment by johnaohara** [comment]
> > If we don't we are locking an instance of Horreum to only accept Tests exported from another Horreum instance that has exported the same TestExport entity.
> 
> I think that's a fair limitation actually :)
> 
> > Adding fields might be fine, but if we remove required fields, change their type, or change the relationship between the test and other entities then it will break.
> 
> Yes, and as I answered above it's a fair limitation imo, but I see your point.
> 
> >

Ok, let us flip it round another way.. We have so far said that the REST API needs to be stable between major versions. What we are now saying is that the REST API **AND** the DB schema needs to stay stable for a major version, if we make a change to the TestExport entity graph, we need to cut a new "major" version.

**Comment by johnaohara** [comment]
> Yes, I also see that changing in the not too distant future. What we then will end up with is a lot of parsing logic to support "old" versions of Horreum none might use...
I need to think about this..

Parsing logic does not need to be complicated, it could be something like:

```
    public void importTest(ObjectNode newTest){
        if ( !newTest.has("version"))
            throw new BadRequestException("Missing field 'version' from the test definition");
        
        String version = newTest.get("version").asText();
        
        switch (version):
            case "1.0":
                TestExport_v1_0 test = mapper.treeToValue(newTest, TestExport_v1.class);
                break;
            case "1.1":
                TestExport_v1_1 test = mapper.treeToValue(newTest, TestExport_v2.class);
                break;
            default:
                throw new BadRequestException("Unknown version: " + version);
        }
```
The might be custom Mapping code, but we could have a base abstract Mapping for a major version, and the subclasses could override only where there has been minor changes.

**Comment by johnaohara** [comment]
@stalep I am not suggesting that we implement versioning *now*, but right now what we can do is change;

```void importTest(TestExport testExport);```

to 

```
void importTest(ObjectNode testExport);

    TestExport testExportObj = mapper.treeToValue(testExport, TestExport.class);
```

With the API configured as;


```
   @GET
   @Path("{id}/export")
   @APIResponseSchema(value = TestExport.class,
           responseDescription = "A Test defintion formatted as json",
           responseCode = "200")
   ObjectNode export(@PathParam("id") int testId);

...

   @POST
   @Path("import")
   @APIResponse(responseCode = "204", description = "Import a new test")
   @RequestBody(content = @Content( mediaType = MediaType.APPLICATION_JSON,
           schema = @Schema( type = SchemaType.STRING, implementation = TestExport.class)) )
   @Operation(description="Import a previously exported Test")
   void importTest(ObjectNode testExport);
```

This would not lock the API into a particular version and we get all the benefits of type safety and auto parsing

**Comment by stalep** [comment]
> > Currently Experiments&/Comparisons are not correctly imported.
> 
> Please can you open an issue to track this

it's the issue this pr is for, #860

**Comment by johnaohara** [comment]
> > > Currently Experiments&/Comparisons are not correctly imported.
> > 
> > 
> > Please can you open an issue to track this
> 
> Ah, you might have misunderstood. Now in main, experiments&comparisons are not working correctly. #860 is tracking that. With this pr it all works (I think).

Ah, sorry, I interpreted  `Currently Experiments&/Comparisons are not correctly imported.` as it was not working with this PR so was asking to open an issue so we could fix that. but if this PR fixes that problem, then please ignore :)

**Comment by johnaohara** [comment]
> Just one last thing, what's the benefit of returning ObjectNode vs TestExport. I see the benefit of importing ObjectNode though.

Good point, I think returning TestExport would be better.

File: horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java (hunk 3 above)
Comment (stalep) [reply horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java:333]: Hm, it's decrypted when we export. Isn't that good enough?
We'll encrypt it when we import.

Comment (stalep) [reply horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java:333]: On only support secrets when we update Tests?

Comment (johnaohara) [reply horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java:333]: > Hm, we could also export some sort of "id" of the Horreum instance... If it doesn't match we set the secret to null?

How do users know if there secrets have been imported or not?  If we set it to null, we need to inform them that they need to configure the secrets again.

Comment (johnaohara) [reply horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java:333]: That could be in a pop-up box, or it could be in the docs, but users need to know the limitations, how to check and what to do

File: docs/site/content/en/docs/Tasks/import-export/index.md [hunk 4]
```diff
@@ -0,0 +1,64 @@
...
+
+## TestSchema
+
+The export object for Schemas is called [SchemaExport](https://github.com/Hyperfoil/Horreum/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/SchemaExport.java) and contains other fields in addition to what's defined in [Schema](https://github.com/Hyperfoil/Horreum/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Schema.java). This includes, labels, extractors and transformers. This is to simplify the import/export experience and make sure that all the data related to a Schema has a single entrypoint with regards to import and export.
```
johnaohara, line 25: For User docs we should focus on what the user sees, not implementation detail. We can describe the JSON structure (or better still link to the OpenAPI spec).

If we are talking about concepts `Test` / `Actions` etc, that should link to those descriptions in the User guide (i.e. https://horreum.hyperfoil.io/docs/concepts/core-concepts/), not the source code

docs/site/content/en/docs/Tasks/import-export/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Import Schemas
```
johnaohara, line 27: I think we need screen shots from the UI as well

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/SchemaExport.java
johnaohara, line 1: This should include openApi annotations

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/TestExport.java
johnaohara, line 1: The fields should be annotated with OpenApi annotations to explain the extra fields, and provide examples

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/TestService.java
```diff
@@ -228,18 +230,18 @@ List<ExportedLabelValues> listLabelValues(@PathParam("id") int testId,
...
 
    @POST
    @Path("import")
-   @APIResponse(responseCode = "204", description = "Import a new test")
+   @APIResponse(responseCode = "204", description = "Import a new Test or update an existing Test")
    @RequestBody(content = @Content( mediaType = MediaType.APPLICATION_JSON,
-           schema = @Schema( type = SchemaType.STRING, implementation = String.class)) )
-   @Operation(description="Import a previously exported Test")
-   void importTest( String testConfig);
+           schema = @Schema( type = SchemaType.STRING, implementation = TestExport.class)) )
```
johnaohara, line 242: The OpenApi viewer is displaying the input payload as "string". This should be a JSON representation of `TestExport`

image

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/SchemaService.java
```diff
@@ -239,19 +241,20 @@ int addOrUpdateTransformer(@PathParam("schemaId") int schemaId,
...
    @POST
    @Path("import")
    @Consumes(MediaType.APPLICATION_JSON)
+   @APIResponse(responseCode = "204", description = "Import a new Schema or update an existing Schema")
    @RequestBody(content = @Content( mediaType = MediaType.APPLICATION_JSON,
            schema = @org.eclipse.microprofile.openapi.annotations.media.Schema(
-                   type = SchemaType.STRING, implementation = String.class)) )
-   @Operation(description="Import an previously exported Schema")
-   void importSchema(String config);
+                   type = SchemaType.STRING, implementation = SchemaExport.class)) )
```
johnaohara, line 255: The OpenApi viewer is displaying the input payload as "string". This should be a JSON representation of SchemaExport

Screenshot from 2024-02-06 06-25-32

horreum-web/package-lock.json
johnaohara, line 1: package-lock.json hits us again :( 

can you reformat the file so that we check in the diffs. 

We *really* need to get on top of https://github.com/Hyperfoil/Horreum/discussions/1245

johnaohara (changes requested):
Code looks great, I think just some things to clean up with the docs and the openAPI spec, and ofc `package-lock.json` :(

horreum-web/package-lock.json
johnaohara: I think this will resolve the issue for us: https://github.com/Hyperfoil/Horreum/pull/1270
When that is merged, you should be able to rebase and only the diffs will be committed

stalep: with rebasing there are no changes to package-lock

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/SchemaService.java
```diff
@@ -239,19 +241,19 @@ int addOrUpdateTransformer(@PathParam("schemaId") int schemaId,
...
 
    @POST
    @Path("import")
    @Consumes(MediaType.APPLICATION_JSON)
-   @RequestBody(content = @Content( mediaType = MediaType.APPLICATION_JSON,
-           schema = @org.eclipse.microprofile.openapi.annotations.media.Schema(
-                   type = SchemaType.STRING, implementation = String.class)) )
-   @Operation(description="Import an previously exported Schema")
-   void importSchema(String config);
+   @APIResponse(responseCode = "204", description = "Import a new Schema or update an existing Schema")
```
johnaohara, line 252: The response code should be "201" : https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/TestService.java
```diff
@@ -228,18 +230,20 @@ List<ExportedLabelValues> listLabelValues(@PathParam("id") int testId,
...
 
    @POST
    @Path("import")
-   @APIResponse(responseCode = "204", description = "Import a new test")
-   @RequestBody(content = @Content( mediaType = MediaType.APPLICATION_JSON,
-           schema = @Schema( type = SchemaType.STRING, implementation = String.class)) )
-   @Operation(description="Import a previously exported Test")
-   void importTest( String testConfig);
+   @Consumes(MediaType.APPLICATION_JSON)
+   @APIResponse(responseCode = "204", description = "Import a new Test or update an existing Test")
```
johnaohara, line 242: This response code should be "201": https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

johnaohara (approved):
LGTM

johnaohara:
Merged. Thanks a lot for your contribution

PR #1213, opened 2024-01-30
upgrade smallrye-openapi to 3.9

stalep:
The upgrade causes the ordering of some argumets for the generated api which required some changes to tsx files

### DIFF
--- a/horreum-web/src/domain/schemas/SchemaList.tsx
+++ b/horreum-web/src/domain/schemas/SchemaList.tsx
@@ -43,7 +43,7 @@ export default function SchemaList() {
     const reloadSchemas = () => {
         setLoading(true)
         schemaApi
-            .list('Ascending', pagination.perPage, pagination.page - 1)
+            .list(pagination.perPage, pagination.page - 1, "", SortDirection.Ascending)
             .then((result) => {
                 setSchemas(result.schemas)
                 setSchemaCount(result.count)
@@ -96,7 +96,7 @@ export default function SchemaList() {
                 Cell: arg => {
                     const changeAccess = useChangeAccess({
                         onAccessUpdate: (id, owner, access) => {
-                            return schemaApi.updateAccess(id, access, owner).then(
+                            return schemaApi.updateAccess(id, owner, access).then(
                                 () => noop(),
                                 error => alerting.dispatchError(error, "SCHEMA_UPDATE", "Failed to update schema access.")
                             ).then(() => reloadSchemas)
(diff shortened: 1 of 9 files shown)


horreum-web/src/domain/schemas/SchemaList.tsx
```diff
@@ -33,22 +33,22 @@ export default function SchemaList() {
...
             setSchemaCount(newCount)
         }
     }
 
     const reloadSchemas = () => {
         setLoading(true)
         schemaApi
-            .list('Ascending', pagination.perPage, pagination.page - 1)
-            .then((result) => {
+            .list(pagination.perPage, pagination.page - 1, "name")
```
stalep, line 46: Looking at the api the third argument is sort by column afaik, not sort direction. Changed it to name, but would like another opinion on it as well.

whitingjr: lgtm, in the situation the third argument is empty then `name` is set as the assigned [value](https://github.com/Hyperfoil/Horreum/blob/master/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java#L234)

whitingjr:
@stalep I have a couple of issues with this PR. Only the 2nd needs any changes now.
1. please don't bundle in formatting changes, next time create a separate pull request dedicated too formatting.
2. Missing sample Tests. Logging in and after loading sample data provided by this script [example-configuration.sh](https://github.com/Hyperfoil/Horreum/blob/master/infra-legacy/example-configuration.sh) it is expected to see two Tests. But non are displayed with `6802b3b4`. 
missing-tests Whereas `28b1fcf9` displays both. Can you fix this problem ?

stalep:
Thanks for the review @whitingjr !
I've rebased now reverting the formatting changes, unfortunately lazyvim has it enabled by default.
I can also confirm that the tests are not showing, not sure why yet. Looking into it

stalep:
Got some help from @johnaohara and the tests should be listed now.

stalep:
Idk why the CI fails...

johnaohara:
This is a situation where package-lock.json is messed up: https://github.com/Hyperfoil/Horreum/pull/1213/files#diff-7bd5bcb1bbc7060dbc4c33ca2164d862273f7518582df29baefca7f356f8daaa

the entire file has been reformatted, so the scope of the diff is the entire file.

stalep:
Hm, I'm only running 'mvn clean install' never npm directly if I can avoid id.

johnaohara:
@stalep please can you rebase on top of master branch, the package-lock.json issue should be fixed now

johnaohara (approved):

PR #1227, opened 2024-01-31
Implementation of using BarValuesChart to display View Component values.

### LINKED ISSUE #744: Bar charts missing from "Views" dataset comparison view
The table for the "labels" tab embed bar charts in the numeric metrics

Screenshot from 2023-10-12 17-26-43

But this feature is missing from a table built from a view

Screenshot from 2023-10-12 17-26-51

whitingjr:
## Fixes Issue

fixes #744 

## Changes proposed

[x] - Render the component values in a table

- [x] All existing tests passed.

### DIFF
--- a/horreum-web/src/domain/runs/DatasetComparison.tsx
+++ b/horreum-web/src/domain/runs/DatasetComparison.tsx
@@ -8,7 +8,7 @@ import {
     EmptyState,
     EmptyStateBody,
     PageSection,
-    Spinner,
+    Spinner, Tooltip,
 } from "@patternfly/react-core"
 import { expandable, ICell, IRow, Table, TableHeader, TableBody } from "@patternfly/react-table"
 import { NavLink } from "react-router-dom"
@@ -35,8 +35,8 @@ type Ds = {
 export default function DatasetComparison() {
     const { alerting } = useContext(AppContext) as AppContextType;
     window.document.title = "Dataset comparison: Horreum"
-    
-    
+
+
     const params = new URLSearchParams(window.location.search)
     const testId = parseInt(params.get("testId") || "-1")
     const [views, setViews] = useState<View[]>([])
@@ -186,6 +186,7 @@ function LabelsComparison({headers, datasets, alerting}: LabelsComparisonProps) 
             </ActionGroup>
             <div ref={componentRef}>
                 <Table
+                    caption="Label(s) Comparison"
                     aria-label="Label comparison"
                     variant="compact"
                     cells={headers}
@@ -213,28 +214,53 @@ type ViewComparisonProps = {
 
 function ViewComparison({headers, view, datasets, alerting}: ViewComparisonProps) {
     const [loading, setLoading] = useState(false)
-    const [rows, setRows] = useState<IRow[]>()
+    const [rows, setRows] = useState<IRow[]>([])
     const token = useSelector(tokenSelector)
+
     useEffect(() => {
         setLoading(true)
         Promise.all(datasets.map(ds => datasetApi.getSummary(ds.id, view.id)))
             .then(
                 summaries => {
-                    setRows(
-                        view.components.map(vc => ({
-                            cells: [
-                                vc.headerName,
-                                ...summaries.map(summary => {
-                                    const render = renderValue(
-                                        vc.render,
-                                        vc.labels.length == 1 ? vc.labels[0] : undefined,
-                                        token
-                                    )
-                                    return render(summary.view?.[vc.id], summary)
-                                }),
-                            ],
-                        }))
-                    )
+                    const rowsData: any[][] = view.components.map(vc => [
+                        vc.headerName,
+                        ...summaries.map(summary => {
+                            const render = renderValue(vc.render, vc.labels.length === 1 ? vc.labels[0] : undefined, token);
+                            return render(summary.view?.[vc.id], summary)
+                        })
+                    ])
+                    const allRows: IRow[] = []
+
+                    rowsData.forEach((rd, index) => {
+                        const isNum = rd.every((data, rdIndex ) => rdIndex === 0 || typeof rd[rdIndex] === 'number' )
+                        allRows.push({
+                            isOpen: isNum ? false: undefined,
+                            cells:  rd.map( (r, index) => ({
+                                    title: typeof r === "object" ? JSON.stringify(r) : r
+                                })
+                            )
+                        })
+                        if (isNum){
+                            allRows.push({
+                                parent: allRows.length - 1,
+                                cells: [
+                                    "",
+                                    {
+                                        title: (
+                                            <BarValuesChart
+                                                values={rd.slice(1)}
+                                                legend={summaries.map(summary => `${summary.runId}/${summary.ordinal + 1}`)}
+                                            />
+                                        ) ,
+                                        props: {
+                                            colSpan: summaries.length
+                                        }
+                                    },
+                                ]
+                            })
+                        }
+                    })
+                    setRows(allRows);
                 },
                 e => alerting.dispatchError( e, "FETCH_VIEW", "Failed to fetch view for one of datasets.")
             )
@@ -254,7 +280,15 @@ function ViewComparison({headers, view, datasets, alerting}: ViewComparisonProps
                 <PrintButton printRef={componentRef} />
             </ActionGroup>
             <div ref={componentRef}>
-                <Table aria-label="View comparison" variant="compact" cells={headers} rows={rows}>
+                <Table
+                    caption="View Component(s) Comparison"
+                    aria-label="View comparison" variant="compact" cells={headers} rows={rows}
+                    isExpandable={true}
+                    onCollapse={(_, rowIndex, isOpen) => {
+                        rows[rowIndex].isOpen = isOpen
+                        setRows([...rows])
+                    }}
+                >
                     <TableHeader />
                     <TableBody />
                 </Table>


willr3:
Do you have a screenshot of how this changes the appearance or a link I can use with my local dev instance to see the change too?

johnaohara:
Comparing the current 0.11.x branch;

Screenshot from 2024-02-05 20-51-13

 against this PR;

Screenshot from 2024-02-05 20-51-20

It appears the dataset comparison view is no longer rendering String values, and is attempting to plot charts when there is no numeric data

horreum-web/src/domain/runs/DatasetComparison.tsx
```diff
@@ -213,28 +210,71 @@ type ViewComparisonProps = {
...
-                        }))
-                    )
+                    const rows: any[][] = []
+                    summaries.forEach((summary, summary_index) => {
+                        let active_view = summary.view
+                        if (active_view != undefined) {
+                            view.components.forEach((component, component_index) => {
+                                let row = rows.find(r => r[0] === component.headerName)
+                                const pos = component_index + 1
+                                // @ts-ignore
```
johnaohara, line 226: Please can we fix the TS errors rather than ignoring them

horreum-web/src/domain/runs/DatasetComparison.tsx
```diff
@@ -213,28 +210,71 @@ type ViewComparisonProps = {
...
+                                } else {
+                                    row[row.length] = val
+                                }
+                            })
+                        }
+                    })
+                    rows.sort((r1, r2) => r1[0].localeCompare(r2[0]))
+                    const renderRows: IRow[] = []
+                    rows.forEach(row => {
+                        const numeric = row.every((value, i) => i === 0 || typeof value === "number")
```
johnaohara, line 250: This appears to be resolving to true, regardless of content, as every line has an expandable chart

horreum-web/src/domain/runs/DatasetComparison.tsx
```diff
@@ -213,28 +210,71 @@ type ViewComparisonProps = {
...
+                        if (active_view != undefined) {
+                            view.components.forEach((component, component_index) => {
+                                let row = rows.find(r => r[0] === component.headerName)
+                                const pos = component_index + 1
+                                // @ts-ignore
+                                const active_component = active_view[pos]
+                                const ob = JSON.stringify( active_component )
+                                let beginning = null
+                                let val = null
+                                if ("\"" === ob.substring(ob.length -2, ob.length - 1)) {
```
johnaohara, line 231: Please can you explain why we need to create a Json String from a Number Array and then perform String manipulation on the result? Why can we not manipulate the objects directly?

horreum-web/src/domain/runs/DatasetComparison.tsx
```diff
@@ -213,28 +210,71 @@ type ViewComparisonProps = {
...
+                                    beginning = ob.lastIndexOf(":", ob.length - 2)
+                                    val = ob.substring( beginning + 2, ob.length - 2)
+                                } else {
+                                    beginning = ob.lastIndexOf(":", ob.length - 1)
+                                    val = Number.parseFloat( ob.substring( beginning +1, ob.length - 1) )
```
johnaohara, line 236: `ViewComponent` works slightly differently to `LabelValues` in that `ViewComponent` can have a value and a render JS function, therefore the `val` displayed to the user is not just a string or a float.

Previously there were calls to `renderValue` to create a render function `render()` that the was called to generate the value displayed to the user.

We need to retain this behaviour so that the data displayed in the UI is consistent

johnaohara (changes requested):
Please see comments. 

We still need to be able to render the View Components values correctly in the Dataset Comparison view

whitingjr:
> Comparing the current 0.11.x branch;
> 
> Screenshot from 2024-02-05 20-51-13
> 
> against this PR;
> 
> Screenshot from 2024-02-05 20-51-20
> 
> It appears the dataset comparison view is no longer rendering String values, and is attempting to plot charts when there is no numeric data

This shows the Dataset view working using example data.
example-default-view

 I need to test using same data as you. Where are you getting your Dataset data from ?

johnaohara:
> > Comparing the current 0.11.x branch;
> > Screenshot from 2024-02-05 20-51-13
> > against this PR;
> > Screenshot from 2024-02-05 20-51-20
> > It appears the dataset comparison view is no longer rendering String values, and is attempting to plot charts when there is no numeric data
> 
> This shows the Dataset view working using example data. example-default-view
> 
> I need to test using same data as you. Where are you getting your Dataset data from ?

Not a public dataset, but if you want a reference to an internal dataset, you can take a look at;

https://example.com/dataset/comparison?testId=262&ds=133265_26488_0&ds=133267_26490_0#labels

horreum-web/src/domain/runs/DatasetComparison.tsx
whitingjr: because what you assume to be an object is implemented as [a string type](https://github.com/Hyperfoil/Horreum/blob/master/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/DatasetService.java#L133).
Converting it to an object would be the answer. But accessing the object property returns `undefined`.

johnaohara: If we need an object we can call `JSON.parse()` and cast to an any type. calling `JSON.stringify( )` on a string and then performing string manipulation is brittle and will break.

Accessing views by position active_view[pos] that is also fragile and likely to break

johnaohara:
Another example of a more complex view is test 62 (internally)

One `ViewComponent` for the Default view for that test is;

```
{
	"6": {
		"id": 242945,
		"headerOrder": 6,
		"headerName": "Mean user CPU",
		"labels": [
			"Mean user CPU"
		],
		"render": "(value, run, token) => new Number( value.reduce((a, b) => { if (!a) { a = 0 } if (!b) { b = 0 } return a + b; }, 0) / value.length ).toPrecision(3)"
	}
}
```

And the `View` in DatasetSummary is;

```
{
	"242945": {
		"Mean user CPU": [
			0.878,
			2.34,
			0.926,
			1.554,
			1.794,
			2.804,
			...
			1.63,
			1.246,
			2.92,
			0.776,
			1.528
		]
	}
}
```

The Label `Mean user CPU` should be rendered with the function `(value, run, token) => new Number( value.reduce((a, b) => { if (!a) { a = 0 } if (!b) { b = 0 } return a + b; }, 0) / value.length ).toPrecision(3)`

To provide a table like;

image

whitingjr:
> > > Comparing the current 0.11.x branch;
> > > Screenshot from 2024-02-05 20-51-13
> > > against this PR;
> > > Screenshot from 2024-02-05 20-51-20
> > > It appears the dataset comparison view is no longer rendering String values, and is attempting to plot charts when there is no numeric data
> > 
> > 
> > This shows the Dataset view working using example data.
> > I need to test using same data as you. Where are you getting your Dataset data from ?
> 
> Not a public dataset, but if you want a reference to an internal dataset, you can take a look at;
> 
> https://example.com/dataset/comparison?testId=262&ds=133265_26488_0&ds=133267_26490_0#labels

I can't recreate what you are seeing. Instead I don't have any data being displayed using the internal dataset. I will need to investigate further.

empty

whitingjr:
> > Comparing the current 0.11.x branch;
> > Screenshot from 2024-02-05 20-51-13
> > against this PR;
> > Screenshot from 2024-02-05 20-51-20
> > It appears the dataset comparison view is no longer rendering String values, and is attempting to plot charts when there is no numeric data
> 
> This shows the Dataset view working using example data. example-default-view
> 
> I need to test using same data as you. Where are you getting your Dataset data from ?

This has been fixed and is now being displayed as expected.
working-Default-view-test-262

horreum-web/src/domain/runs/DatasetComparison.tsx
```diff
@@ -213,28 +214,54 @@ type ViewComparisonProps = {
...
+                        return rowCells;
+                    });
+                    const allRows: IRow[] = []
+
+                    rowsData.forEach((rd, index) => {
+                        const isNum = typeof rd[rd.length -1] === 'number'
+                        allRows.push({
+                            isOpen: isNum ? false: undefined,
+                            cells:  rd.map( (r, index) => ({
+                                    title: r
```
willr3, line 240: I think `r` could be a json object which probably cannot be directly injected into the cell.

horreum-web/src/domain/runs/DatasetComparison.tsx
```diff
@@ -213,28 +214,54 @@ type ViewComparisonProps = {
...
-                        }))
-                    )
+                    const rowsData: any[][] = view.components.map(vc => {
+                        const rowCells = [vc.headerName];
+                        summaries.forEach(summary => {
+                            const render = renderValue(vc.render, vc.labels.length === 1 ? vc.labels[0] : undefined, token);
+                            rowCells.push(render(summary.view?.[vc.id], summary));
+                        });
+                        return rowCells;
+                    });
```
willr3, line 232: Is there a reason you removed the spread notation `[vc.headerName, ...summaries.map(` that was used before?
I think that would be easier to read but I'm curious if you found explicit iteration and mutation easier to read

whitingjr: So while it might be pleasing to the eye I find it more straightforward to pass the function to render function. Also I need to capture this into a intermediate datastructure for further processing to detect numeric rows. Does that make sense why I didn't use spread notation ?

willr3:
```javascript
const rowsData: any[][] = view.components.map(vc => {
                        const rowCells = [vc.headerName];
                        summaries.forEach(summary => {
                            const render = renderValue(vc.render, vc.labels.length === 1 ? vc.labels[0] : undefined, token);
                            rowCells.push(render(summary.view?.[vc.id], summary));
                        });
                        return rowCells;
                    });
```
could be more succinctly expressed as 
```javascript
const rowsData: any[][] = view.components.map(vc => [
  vc.headerName,
  ...sumaries.map(summary => {
    const render = renderValue(vc.render, vc.labels.length === 1 ? vc.labels[0] : undefined, token);
    return render(summary.view?.[vc.id], summary)
  })
])  
```
They both produce the same `rowsData` content. I was curious if you had a preference for the explicit iteration with `forEach` and `push`. If you don't then would you please use the more succinct expression? I think it makes the code easier to read.

horreum-web/src/domain/runs/DatasetComparison.tsx
```diff
@@ -213,28 +214,54 @@ type ViewComparisonProps = {
...
+                    const allRows: IRow[] = []
+
+                    rowsData.forEach((rd, index) => {
+                        const isNum = typeof rd[rd.length -1] === 'number'
```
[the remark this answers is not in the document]
willr3: I think this is a bug, the `const isNum` is only checking the last entry in `rd` and assuming all values are numeric if the last value is numeric. It is very likely that one dataset could be missing the extracted value. The linked `LabelComparison` checked every value (aside from the header) to ensure all values are numbers.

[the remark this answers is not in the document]
whitingjr: changed to check all values of rd.

horreum-web/src/domain/runs/DatasetComparison.tsx
```diff
@@ -213,28 +214,53 @@ type ViewComparisonProps = {
...
+                        })
+                    ])
+                    const allRows: IRow[] = []
+
+                    rowsData.forEach((rd, index) => {
+                        const isNum = rd.every((data, rdIndex ) => rdIndex === 0 || typeof rd[rdIndex] === 'number' )
+                        allRows.push({
+                            isOpen: isNum ? false: undefined,
+                            cells:  rd.map( (r, index) => ({
+                                    title: typeof r === "object" ? JSON.stringify(r) : r
```
willr3, line 239: Thank you for sanitizing this but the more I think about it I think the value of `r` is the `v.headerName` or the output of the `render(...)` and I'm not sure if that is normally placed under the `title` for a `cell`. I'm going to read up on the `Table` `rows` attribute before we merge this

willr3: https://www.patternfly.org/components/table/react-deprecated#table-rows
`IRowCell.title` is where the cell content is placed so this looks good to me.

PR #1254, opened 2024-02-04
Remove moment library from UI

shivam-sharma7:


horreum-web/src/utils.ts
```diff
@@ -152,3 +152,29 @@ export function fingerprintToString(fingerprint: unknown) {
     }
     return JSON.stringify(fingerprint);
 }
+
+export function timeFromNow(date: number) {
```
willr3, line 156: What do you think about using the existing `luxon` library and `Duration` to perform the calculations?
Something like what they were doing in this comment but without the `Intl` dependency
https://github.com/moment/luxon/issues/274#issuecomment-729126515

shivam-sharma7: @willr3 Our goal is to handle `lock` file and any dependency that is not providing significant or  less use so we don't need conflicts any more.

willr3: I thought we were removing `moment` to decrease the build size of the deployed javascript. In either case, I think the proposed change will result in `1 years ago` instead of `1 year ago` (same issue with other increments)

willr3:
I do not think we should write our own formatter. The recommendation from Patternfly5 is to use the Day.js dependency
https://www.patternfly.org/ux-writing/numerics#date-and-time-formats

shivam-sharma7:
@willr3 We are already rely on too many dependencies that are not giving significant. Then how can we manage our dependecies and What am I thinking that we don't rely on too many dependencies.

horreum-web/src/domain/runs/components.tsx
```diff
@@ -74,7 +74,7 @@ export const ExecutionTime = (timestamps: StartStop) => (
             </table>
         }
     >
-        <span>{moment(toEpochMillis(timestamps.stop)).fromNow()}</span>
+        <span>{timeFromNow(toEpochMillis(timestamps.stop))}</span>
```
shivam-sharma7, line 77:
```javaScript
<span>{dayjs(toEpochMillis(timestamps.stop)).fromNow()}</span>
```
@willr3 I think you want to suggest above changes with day.js

PR #1264, opened 2024-02-05
Always fetch roles from backend :  Fixes #1255

### LINKED ISSUE #1255: UI fetch keycloak roles through the backend
At the moment the UI fetches the roles from the OIDC login. These should be fetched from the backend instead, even if in turn  the roles are fetch from keycloak too, to allow a  transition to the roles defined in Horreum.

barreiro:
Fixes #1255


johnaohara (approved):

johnaohara:
Merged. Thanks a lot for your contribution

PR #1288, opened 2024-02-07
Fix devservices with DB backup

barreiro:
- Propagate credentials from .env file 
- Remove the hard-coded IP address and replace with the defined network alias

### DIFF
--- a/infra/horreum-infra-common/src/main/java/io/hyperfoil/tools/horreum/infra/common/resources/KeycloakResource.java
+++ b/infra/horreum-infra-common/src/main/java/io/hyperfoil/tools/horreum/infra/common/resources/KeycloakResource.java
@@ -39,9 +39,6 @@ public class KeycloakResource implements ResourceLifecycleManager {
                 throw new RuntimeException("Arguments did not contain Keycloak image");
             }
 
-            final String JDBC_URL = initArgs.get("quarkus.datasource.jdbc.url.internal");
-            final String DB_PORT = initArgs.get("postgres.container.port");
-
             final String KEYCLOAK_IMAGE = initArgs.get(HORREUM_DEV_KEYCLOAK_IMAGE);
 
             networkAlias = initArgs.get(HORREUM_DEV_KEYCLOAK_NETWORK_ALIAS);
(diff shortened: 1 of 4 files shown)


infra/horreum-infra-common/src/main/java/io/hyperfoil/tools/horreum/infra/common/resources/KeycloakResource.java
```diff
@@ -100,7 +97,7 @@ public void init(Map<String, String> initArgs) {
                         .withEnv("KC_DB_URL_HOST", "")
                         .withEnv("KC_HOSTNAME_STRICT", "false")
                         .withEnv("KC_HOSTNAME", "localhost")
-                        .withEnv("KC_DB_URL", "jdbc:postgresql://172.17.0.1:" + DB_PORT + "/keycloak")
+                        .withEnv("KC_DB_URL", "jdbc:postgresql://" + initArgs.get(HORREUM_DEV_POSTGRES_NETWORK_ALIAS) + ":5432/keycloak")
```
johnaohara, line 100: Can we leave the `DB_PORT` as configurable? pple may spin up multiple instances on horreum in dev mode on the default network

johnaohara:
Credentials should already be overwritten by an `.env` file, is that not working?

One thing I have noticed is that we have 
`%dev.horreum.dev-services.keycloak.admin-username=`

and

`%dev.quarkus.keycloak.admin-client.username`

We should probably drop the `horreum.dev-services.keycloak.admin-username` / `horreum.dev-services.keycloak.admin-password` config altogether and just use the inbuilt `quarkus.keycloak.admin-client` credentials. will remove duplication of the same creds

barreiro:
> Credentials should already be overwritten by an .env file, is that not working?

Yes, but it's the other way around: we need to set those when there is no .env file.

> We should probably drop the `horreum.dev-services.keycloak.admin-username` / `horreum.dev-services.keycloak.admin-password` config altogether and just use the inbuilt `quarkus.keycloak.admin-client` credentials. will remove duplication of the same creds

Perhaps, in that case we would need to update the `.env` files to use those properties instead.

johnaohara:
> Yes, but it's the other way around: we need to set those when there is no .env file. 

When have you had to set them? it should work without an .env file and build an empty dev environment

barreiro:
updated the PR to set `quarkus.keycloak.admin-client` credentials from `horreum.dev-services.keycloak` properties. These can come either from defaults or the `.env` file.

infra/horreum-infra-common/src/main/java/io/hyperfoil/tools/horreum/infra/common/resources/KeycloakResource.java
johnaohara: This is  going to break my workflow where I have multiple instances running in dev mode

PR #1290, opened 2024-02-07
Horreum UI refactoring

johnaohara:
Horreum UI refactoring.

In summary, here the most important changes:
- [x] Main toolbar refactored as left-sidebar
- [x] Test-specific toolbar improved
   - [x] Runs and datasets showed as inner tabs
   - [x] Changes and reports are no more standalone pages, they are under a specific test
   - [x] "General" tab renamed as "Settings"
- [x] Folders navigation refactored using `breadcrumb`

### DIFF
--- a/horreum-web/src/domain/alerting/Changes.tsx
+++ b/horreum-web/src/domain/alerting/Changes.tsx
@@ -1,7 +1,6 @@
 import {useState, useEffect, useMemo, useCallback, useContext} from "react"
-import { useSelector } from "react-redux"
 import { ChangesTabs } from "./ChangeTable"
-import TestSelect, { SelectedTest } from "../../components/TestSelect"
+import { SelectedTest } from "../../components/TestSelect"
 import LabelsSelect, { convertLabels } from "../../components/LabelsSelect"
 import PanelChart from "./PanelChart"
 import { fingerprintToString, formatDate } from "../../utils"
@@ -18,11 +17,9 @@ import {
 } from "../../api"
 
 import {
-	Button,
 	Card,
 	CardBody,
 	CardHeader,
-	ClipboardCopy,
 	DataList,
 	DataListItem,
 	DataListItemRow,
@@ -31,20 +28,20 @@ import {
 	DatePicker,
 	EmptyState,
 	EmptyStateBody,
-	Modal,
-	PageSection,
-	Spinner, EmptyStateHeader, EmptyStateFooter,	
+	Spinner, EmptyStateHeader, EmptyStateFooter,
 } from '@patternfly/react-core';
 import {
 	Select,
 	SelectOption,
 	SelectOptionObject
 } from '@patternfly/react-core/deprecated';
-import { NavLink, useNavigate } from "react-router-dom"
+import { useNavigate } from "react-router-dom"
 import {AppContext} from "../../context/appContext";
 import {AppContextType} from "../../context/@types/appContextTypes";
+import {useSelector} from "react-redux";
 
 type TimespanSelectProps = {
+    value?: number
     onChange(span: number): void
 }
 
@@ -71,7 +68,8 @@ const TimespanSelect = (props: TimespanSelectProps) => {
         ],
         []
     )
-    const [selected, setSelected] = useState(options[4])
+    const defaultOption = options.find((t, _) => { return t.seconds === props.value })
+    const [selected, setSelected] = useState(defaultOption ?? options[4])
     return (
         <Select
             isOpen={isExpanded}
@@ -94,6 +92,7 @@ const TimespanSelect = (props: TimespanSelectProps) => {
 }
 
 type LineTypeSelectProps = {
+    value?: string
     onChange(type: string): void
 }
 
@@ -115,7 +114,8 @@ const LineTypeSelect = (props: LineTypeSelectProps) => {
         ],
         []
     )
-    const [selected, setSelected] = useState(options[1])
+    const defaultOption = options.find((l, _) => { return l.type === props.value })
+    const [selected, setSelected] = useState(defaultOption ?? options[1])
     return (
         <Select
             isOpen={isExpanded}
@@ -204,16 +204,21 @@ export const flattenNode = (arr : Array<any> | undefined) => {
     return nodeObj;
 }
 
+type ChangesProps = {
+    testID: number
+}
 
-export default function Changes() {
+
+export default function Changes(props: ChangesProps) {
     const { alerting } = useContext(AppContext) as AppContextType;
     const navigate = useNavigate()
     const params = new URLSearchParams(location.search)
-    // eslint-disable-next-line
-    const paramTest = useMemo(() => params.get("test") || undefined, [])
-    const paramFingerprint = params.get("fingerprint")
     const teams = useSelector(teamsSelector)
-    const [selectedTest, setSelectedTest] = useState<SelectedTest>()
+    const newTest = { id: props.testID } as SelectedTest;
+    // This could be removed in future as the test id is provided by props therefore it should not change
+    const [selectedTest, setSelectedTest] = useState<SelectedTest>(newTest)
+
+    const paramFingerprint = params.get("fingerprint")
     const [selectedFingerprint, setSelectedFingerprint] = useState<FingerprintValue | undefined>(() => {
         if (!paramFingerprint) {
             return undefined
@@ -224,15 +229,11 @@ export default function Changes() {
             return { ...fingerprint, toString: () => str }
         } catch (e) {
             alerting.dispatchError("Failed to parse fingerprint <code>" + paramFingerprint + "</code>",
-                    "PARSE_FINGERPRINT",
-                    "Fingerprint parsing failed")
+                "PARSE_FINGERPRINT",
+                "Fingerprint parsing failed")
             return undefined
         }
     })
-    const [selectedChange, setSelectedChange] = useState<number>()
-    const [selectedVariable, setSelectedVariable] = useState<number>()
-
-
     const [panels, setPanels] = useState<PanelInfo[]>([])
     const [loadingPanels, setLoadingPanels] = useState(false)
     const [loadingFingerprints, setLoadingFingerprints] = useState(false)
@@ -245,7 +246,7 @@ export default function Changes() {
     const [lineType, setLineType] = useState(params.get("line") || "linear")
 
     const createQuery = (alwaysEndTime: boolean) => {
-        let query = "?test=" + selectedTest
+        let query = ""
         if (selectedFingerprint) {
             query += "&fingerprint=" + encodeURIComponent(JSON.stringify(selectedFingerprint))
         }
@@ -258,14 +259,14 @@ export default function Changes() {
         if (lineType !== "linear") {
             query += "&line=" + lineType
         }
-        return query
+        return "?" + query.replace(/^&/, '');
     }
     useEffect(() => {
         if (!selectedTest) {
             document.title = "Changes | Horreum"
             return
         }
-        document.title = `${selectedTest} | Horreum`
+        document.title = `${selectedTest.id} | Horreum`
         navigate(location.pathname + createQuery(false))
     }, [selectedTest, selectedFingerprint, endTime, timespan, lineType, firstNow, history])
     useEffect(() => {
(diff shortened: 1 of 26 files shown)


willr3:
Screenshot from 2024-02-07 21-04-33

Overall I like the idea of moving items under the associated `Test` and preserving the selected test when navigating between them.

My opinion on changes:
* The `login` button should stay right aligned in the top nav.
* Only 2 (3 with `Admin`) options in the sidebar uses a lot of vertical space. 
* What happened to `Runs`? Users upload `Runs` not `Datasets` so I think `Datasets` are what should go away, not `Runs`.

johnaohara:
>    The login button should stay right aligned in the top 

That was my intention, it is layout issue that needs resolving

>    Only 2 (3 with Admin) options in the sidebar uses a lot of vertical space.

I was thinking about being about to "favourite" tests which would appear under the "Test" menu, this would fill this sidebar up.

It is also possible to collapse the sidebar so you get the full width of the screen by pressing the  button
image

The other option is to put `Tests` / `Schema` / `Admin` back on the top menu and some of the 

>    What happened to Runs? Users upload Runs not Datasets so I think Datasets are what should go away, not Runs.

I was experimenting with simplifying the UI, atm the core abstraction for change detection is dataset. The "Runs" View allows you to compare runs and delete them, but is similar to datasets.

Overall we need to think about what is the best way to visualize the Run/Dataset abstraction, or if we need that abstraction at all

johnaohara:
The "Folder" navigation could move into the navigation bar
image

and we get a lot of space back at the top of the main view

image

instead of what we have right now:

image

lampajr:
## What I did
In the previous commits I have fixed (hopefully) some UI issues that I'll summarize here:
1. Aligned the header toolbar to the right such that the `Log in` is in the top-right corner
image
2. Simplified the `changes` tab query params, removing the weird `http://localhost:8080/test/113/?test=[object%20Object]` test param (which is not really needed as already present in the URI)
image
3. Fixed the Alert box message as it was overflowing when the left sidebar was opened
4. Fixed the reports navigation as it was commented out
   * Here I re-established the same UI that was there before, I simply changed the full URI to contain the `test/:testId` as root.
   * For instance, I changed from `/reports/table/config/:configId` to `/test/:testId/reports/table/config/:configId`
   * Now the `reports` link moves the the user back to the reports tab for that specifc `test`
image

## Open points
In conclusion I think that from this pull request point of view we have still ~2 open points:
* Is the `reports` navigation I re-established fine? Or were we thinking about a different approach?
* Right now the folder selection is just a dropdown list
image
   * Once selected the folder name is not shown in that dropdown button
image
   * Do we want to implement the folder structure as mentioned/showed in this comment https://github.com/Hyperfoil/Horreum/pull/1290#issuecomment-1933469772 ?

lampajr:
Alternatively another option for the Folder structure could be to use the `breadcrumb` approach like:
image
image
image

I still don't know where the `Team Selector` could be placed, as I don't really like it next to the breadcrumb.
Any opinion on this approach?

horreum-web/src/components/SavedTabs.tsx
```diff
@@ -101,7 +101,7 @@ export default function SavedTabs(props: SavedTabsProps) {
                             const childProps = children[activeKey.current].props
                             if ("onSave" in childProps) {
                                 childProps
-                                    .onSave()
+                                    .onSave?.()
```
lampajr, line 104: I don't think this is actually required as `onSave` is a mandatory property, isn't it?
```suggestion
                                    .onSave()
```

horreum-web/src/domain/alerting/Changes.tsx
```diff
@@ -204,16 +205,21 @@ export const flattenNode = (arr : Array<any> | undefined) => {
...
     const navigate = useNavigate()
     const params = new URLSearchParams(location.search)
-    // eslint-disable-next-line
-    const paramTest = useMemo(() => params.get("test") || undefined, [])
-    const paramFingerprint = params.get("fingerprint")
     const teams = useSelector(teamsSelector)
-    const [selectedTest, setSelectedTest] = useState<SelectedTest>()
+    const newTest = { id: props.testID } as SelectedTest;
+    // Do we need this? the test id is provided by props therefore it should not change
+    const [selectedTest, setSelectedTest] = useState<SelectedTest>(newTest)
```
lampajr, line 220: Do we need this? the test id is provided by props therefore it should not change. In fact the function `setSelectedTabs` is not called anywhere

horreum-web/src/domain/tests/AllTests.tsx
```diff
@@ -394,22 +372,20 @@ export default function AllTests() {
...
+
     const loadTests = () => {
-        setLoading(true)
         fetchTestsSummariesByFolder(alerting, SortDirection.Ascending, pagination.folder, pagination.limit, pagination.page, rolesFilter.key, )
-            .then(summary => {setTests(summary.tests?.map(t => mapTestSummaryToTest(t)) || [])
-                if (summary.count) {
-                    setCount(summary.count)
-                }
-            })
-            .finally( () => setLoading(false))
```
lampajr: I see the `loading` management has been removed, this means that `const [loading, setLoading] = useState(false)` would be no longer needed as no more used anywhere here. Was that intentional?

horreum-web/src/domain/tests/AllTests.tsx
```diff
@@ -394,22 +372,25 @@ export default function AllTests() {
...
+
     const loadTests = () => {
-        setLoading(true)
         fetchTestsSummariesByFolder(alerting, SortDirection.Ascending, pagination.folder, pagination.limit, pagination.page, rolesFilter.key, )
-            .then(summary => {setTests(summary.tests?.map(t => mapTestSummaryToTest(t)) || [])
+            .then(summary => {
+                setTests(summary.tests?.map(t => mapTestSummaryToTest(t)) || [])
                 if (summary.count) {
                     setCount(summary.count)
                 }
```
lampajr, line 391: With commit https://github.com/Hyperfoil/Horreum/pull/1290/commits/9781289787ac737f8b749e60680db44e9b1c1a1f I had to re-add this `setCount` call otherwise the pagination would not work as expected (always set to 0)

lampajr:
As discussed added additional commit (https://github.com/Hyperfoil/Horreum/pull/1290/commits/ca3fff3452e872dc0df7da8a40de8db0cf3b96e2) to add folders navigation using breadcrumb:
image

I tried to keep the folder structure, i.e., `Horreum` as root of the tree and then all the folders, an alternative could be to remove that root layer:
image

lampajr:
For better navigation the last commit introduce the original folder where the test resides, e.g.,
image

lampajr:
Since we changed some of the URL paths, e.g., for `changes`, IIUC we should change the URLs we inject into the email notification as well (assuring they will point to the correct resources)

johnaohara:
> 
> I tried to keep the folder structure, i.e., `Horreum` as root of the tree and then all the folders, an alternative could be to remove that root layer: image

I think the drop down with "24" is confusing, it is not obvious what the "24" refers to.  I think using "Test" in the breadcrumb for the dropdown might be more intuitive for users;

Screenshot from 2024-04-09 05-57-05

I would also drop the "<" arrow from Horreum, that way it appears to be a tree, with "Horreum" as the root

johnaohara:
When I click on a "Run" of a "Dataset", the breadcrumb disappears.

I think we should retain the breadcrumb on all views

Screenshot from 2024-04-09 06-03-00

johnaohara:
> Is the `reports` navigation I re-established fine? Or were you thinking about a different approach?
> 
>     * Maybe this is a question more for @johnaohara

@lampajr  What changes were made to the `reports` navigation? I can not see an issue, but was wondering what had changed?

horreum-web/src/domain/alerting/Changes.tsx
johnaohara: I think we can remove this now

lampajr:
> > Is the `reports` navigation I re-established fine? Or were you thinking about a different approach?
> > ```
> > * Maybe this is a question more for @johnaohara
> > ```
> 
> @lampajr What changes were made to the `reports` navigation? I can not see an issue, but was wondering what had changed?

@johnaohara actually nothing changed (it was simply commented out, so I re-established it), the only difference with respect to the `as-is` is the URL that hanged from `/reports/table/config/:configId` to `/test/:testId/reports/table/config/:configId`

> I think the drop down with "24" is confusing, it is not obvious what the "24" refers to. I think using "Test" in the breadcrumb for the dropdown might be more intuitive for users;

I agree, much cleaner and easier to underastand. Working on it.

> When I click on a "Run" of a "Dataset", the breadcrumb disappears.
I think we should retain the breadcrumb on all views

Oh, good catch! I did not notice. Working on this as well.

lampajr:
I was able to keep the breadcrumb on `run` and `dataset` view but unfortunately we don't have information about the folder (this could require some refactoring to pass that information down till the final view).

This is the actual result
image

Removed the number of folders from the dropdown, this is the result as you already showed:
image
image
image

horreum-web/src/components/SavedTabs.tsx
johnaohara: It has been a while since this change was made, but... iirc, i **think** I changed `SavedTabs.tsx` for the "Save" button to be optional, or at least that was the intention.  Not all tabs had a natural state that should be saved when navigating away from them, or I had moved some tabs that handling saving state slightly differently.

horreum-web/src/domain/tests/AllTests.tsx
johnaohara: ok

horreum-web/src/domain/tests/AllTests.tsx
johnaohara: @lampajr is this still an question? the code is flagged as "outdated"

lampajr: yeah it is still valid question as the `loading` state is no more managed, therefore I was wondering if that has been done on purpose or not

johnaohara:
On the `Tests` view, there is the option to filter tests by ownership in a dropdown.

That same dropdown is in the Schema view, but is not currently functional.  I think we should get this wired up

Screenshot from 2024-04-12 06-15-14

horreum-web/src/domain/tests/AllTests.tsx
lampajr: yeah sure

horreum-web/src/components/SavedTabs.tsx
lampajr: yeah makes sense, I checked further and the props could be either `FragmentTabProps` or `SavedTabProps` and the `onSave` method is present only in the latter (where it must be always present). 
Moreover you are checking a couple of lines before if `onSave` is present in the child props so I think it is save to remove that `?.` notation. Anyway just a minor :)

johnaohara: sure, change it to `.onSave()`

horreum-web/src/domain/tests/AllTests.tsx
lampajr: Done

lampajr:
Hi @johnaohara 

I addressed the other minors, especially I re-instated the `setLoading` on `AllTests` page.

> That same dropdown is in the Schema view, but is not currently functional. I think we should get this wired up

Yeah that's something we should do, I checked and it seems this is not supported from the backend therefore it is not about UI changes only. Do you want to do that as part of this PR? Or do you prefer to create a separate issue to manage both backend and frontend?

lampajr:
> If it is not supported in the backend, we should hide the component in the UI and add that in a separate PR

I agree, hidden in this pr and created https://github.com/Hyperfoil/Horreum/issues/1589 for its implementation.
I raised that as bug since it is actually showed in production so users would exepct to work.

johnaohara:
@lampajr please can you change title the PR, squash commits and remove from Draft status?

shivam-sharma7:
@johnaohara will we nee all commentend code, I am asking because its kind of mesh.

johnaohara:
> @johnaohara will we nee all commentend code, I am asking because its kind of mesh.

Not in this instance.  In this PR, the UI is being re-organised as an interim step, but there is more re factoring to be performed to allow for Unit testing etc.  As we perform work to improve the UI components, rather than just moving them around, I think we want more details about the components then

whitingjr:
Introducing 3 clicks to navigation is a backward step imo.
1. click menu to open
2. click Test or Schema
3. click menu to hide
Created #1622 for a feature request to make it use mouse events to open/close.

Issue #1073, opened 2024-01-08
Evaluate Experiments button missing.


whitingjr:
**Describe the bug**

Loading a Dataset to inspect an Experiment outcomes not loading correctly. The Evaluate Experiments button missing as well.

![missing-button-where](https://github.com/Hyperfoil/Horreum/assets/708428/9d5af679-ec60-47ac-a393-41500f956ebe)

The UI displays this message 

![ui-error](https://github.com/Hyperfoil/Horreum/assets/708428/8b075414-1c15-4048-85ec-f6635bababf9)

The server shows these exceptions

```
2024-01-08 10:41:21,235 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-115) HTTP Request to /api/experiment/20/profiles failed, error id: 471f2550-e1eb-486e-a9e8-d0ad40988d3c-471: org.hibernate.InstantiationException: Could not instantiate entity : io.hyperfoil.tools.horreum.entity.ExperimentComparisonDAO
	at org.hibernate.metamodel.internal.EmbeddableInstantiatorPojoStandard.instantiate(EmbeddableInstantiatorPojoStandard.java:79)
	at org.hibernate.sql.results.graph.embeddable.AbstractEmbeddableInitializer.createCompositeInstance(AbstractEmbeddableInitializer.java:347)
	at org.hibernate.sql.results.graph.embeddable.AbstractEmbeddableInitializer.prepareCompositeInstance(AbstractEmbeddableInitializer.java:249)
	at org.hibernate.sql.results.graph.embeddable.AbstractEmbeddableInitializer.initializeInstance(AbstractEmbeddableInitializer.java:188)
	at org.hibernate.sql.results.internal.InitializersList.initializeInstance(InitializersList.java:70)
	at org.hibernate.sql.results.internal.StandardRowReader.coordinateInitializers(StandardRowReader.java:111)
	at org.hibernate.sql.results.internal.StandardRowReader.readRow(StandardRowReader.java:87)
	at org.hibernate.sql.results.spi.ListResultsConsumer.consume(ListResultsConsumer.java:179)
	at org.hibernate.sql.results.spi.ListResultsConsumer.consume(ListResultsConsumer.java:33)
	at org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl.doExecuteQuery(JdbcSelectExecutorStandardImpl.java:361)
[...]

PR #1292, opened 2024-02-07
Added @Schema definitions on JsonNode fields, fixes #1278

### LINKED ISSUE #1278: Fix the OpenAPI spec to generate correct python client
**Describe the bug**

Fix the OpenAPI spec to generate correct python client: see https://github.com/Hyperfoil/horreum-client-python/issues/1#issuecomment-1928116857

stalep:


horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/alerting/ChangeDetection.java
```diff
@@ -11,6 +13,7 @@ public class ChangeDetection {
     public String model;
     @NotNull
     @JsonProperty( required = true )
+    @Schema(implementation = JsonNode.class, type = SchemaType.STRING)
     public JsonNode config;
```
johnaohara, line 17: This property should be an `ObjectNode`,  bugs have occurred in the past when we allow `JsonNode` on the API and objects are pushed into the database as text.  That is what lead to this error: https://github.com/Hyperfoil/Horreum/issues/1073

PR #1300, opened 2024-02-08
fixes #1298 by avoiding redundant keycloak init

### LINKED ISSUE #1298: Dev mode Authentication not working as expected
**Describe the bug**

There are intermittent issues with Dev mode authentication method that switches from oidc to basic auth, which makes it very difficult to develop UI features.

The authentication mode should be configured and not change dynamically 

For example:
image

When this happens I can not test any of the UI that I am working on as the dialog box keeps re-appearing

willr3:


horreum-web/src/keycloak.ts
```diff
@@ -8,8 +8,13 @@ import { keycloakSelector, INIT, STORE_PROFILE, UPDATE_DEFAULT_TEAM, UPDATE_ROLE
 
 export function initKeycloak(state: State) {
     const keycloak = keycloakSelector(state)
-    let oidc: boolean
+    let oidc: boolean = false //defaulting to a value
     let keycloakPromise
+
+    if (keycloak && keycloak.authenticated){
+        return;//don't need to setup keycloak again
```
shivam-sharma7, line 15:
```suggestion
        return; //don't need to setup keycloak again
```
space

shivam-sharma7 (approved):

PR #1317, opened 2024-02-13
fix support for async javascript functions

willr3:
async javascript functions return a Promise that needs to be resolved for Horreum to correctly represent the value. This currently fixes label calcuation and the `Test label calculation` in the schema page but I'm sure there are more places where the javascript evaluation needs to be updated.

We really should consolidate the javascript evaluation but for first we fix the issues


stalep:
@willr3 can you add an issue to this so we can track the fix please?

willr3:
> @willr3 can you add an issue to this so we can track the fix please?

I created #1320 to track the issue

johnaohara:
@willr3 can we move the fetch into another issue/PR?  

This PR is adding additional functionality into a PR to fix an issue.  I think we need to discuss this new feature and what use case we are targetting with it.  We also need to assess the impact it has on security etc, for example the webhooks  feature is restricted to an "approved list" of backend calls, the included fetch feature does not

willr3:
I removed all of the Shim's that added support for fetch

willr3:
@johnaohara is there aything you see in the PR that is probihiting from being merged?

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
```diff
@@ -61,22 +115,99 @@ public void toInstant_nulls() throws UnsupportedEncodingException {
...
+    }
+    @org.junit.jupiter.api.Test
+    public void evaluateOnceJsonKeyAccess() throws JsonProcessingException {
+        Object rtrn = Util.evaluateOnce(
+            """
+                    (input)=>{
+                        return input.foo;
+                    }
+                    """,
+                new ObjectMapper().readTree("{\"foo\":\"bar\"}"), //no input needed
```
lampajr, line 144: I think just a leftover
```suggestion
                new ObjectMapper().readTree("{\"foo\":\"bar\"}"),
```

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
```diff
@@ -559,26 +593,87 @@ static <T> void evaluateWithCombinationFunction(List<T> inputData,
...
+         if(invokeRtrn instanceof Value){
+            Value invokedValue = (Value)invokeRtrn;
+         }
+         if(!rejected.isEmpty()){
+            value = rejected.get(0);
+         }else if(resolved.size() == 1){
+            value = resolved.get(0);
+         }else{ //resolve.size() > 1, this doesn't happen
+
+         }
```
lampajr, line 645: Remove empty `else` stmt (we could still keep the comment `//resolve.size() > 1 doesn't happen` if we want)
```suggestion
         }
```

pom.xml
```diff
@@ -81,7 +81,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <commons.math3.version>3.6.1</commons.math3.version>
-        <graalvm.version>23.0.3</graalvm.version>
+        <graalvm.version>23.0.1</graalvm.version>
```
lampajr, line 84: I might have missed that but why do we need to downgrade the `graalvm.version`?

lampajr (commented):
Hey @willr3, I added just a couple of minors (not blockers fmpov).

Anyway I think we will probably need some automated formatting to avoid format issues :smile:

lampajr:
Another thing that I think it i sworth to mention is that, the `graalvm` that we set here will conflict with the one set as part of the Quarkus upgrade https://github.com/Hyperfoil/Horreum/pull/1571/. Therefore we will probably need some additional effort to adapt this change with the newest version of `graalvm`

pom.xml
willr3: good catch, this was likely a merge conflict I missed / clicked incorrectly

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
willr3: instead of an empty `else` perhaps we should log an error that the resolve violated Horreum's expectation?

willr3:
@lampajr yes, we have to be careful when performing updates to Quarkus or anything that could change graalvm. Horreum needs testing to ensure any graalvm language operates correctly. I now there are breaking changes in newer versions of graal as I've run into them with `yaup`

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
lampajr: yeah makes sense, so if it happens we have logs to track it

horreum-backend/pom.xml
```diff
@@ -40,6 +41,24 @@
...
+            <artifactId>polyglot</artifactId>
+            <version>${graalvm.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.graalvm.polyglot</groupId>
+            <artifactId>js-community</artifactId>
+            <version>${graalvm.version}</version>
+            <type>pom</type>
+            <scope>import</scope>
+        </dependency>-->
```
johnaohara, line 56: Can we remove these comments and handle in https://github.com/Hyperfoil/Horreum/pull/1571? otherwise we are going to get a conflict here

johnaohara (changes requested):

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
willr3: I will update it to include a `log.error` rather than just a comment stating the else branch is unexpected.

pom.xml
willr3: I was not intentionally downgrading. I am either missing a recent commit that bumped the version or I messed up my merge during a rebase

willr3:
I pushed a fix that I believe addressed all reviews. Please let me know what I missed :)

horreum-backend/pom.xml
lampajr, line 1: You need to exclude the following deps
```xml
                <exclusion>
                    <groupId>org.graalvm.regex</groupId>
                    <artifactId>regex</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>org.graalvm.truffle</groupId>
                    <artifactId>truffle-api</artifactId>
                </exclusion>
```

from the `org.graalvm.js:js` dependency in horreum-backend.

horreum-backend/pom.xml
```diff
@@ -40,6 +40,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.graalvm.regex</groupId>
+            <artifactId>regex</artifactId>
+            <version>${graalvm.version}</version>
+        </dependency>
```
lampajr, line 47:
```suggestion
        <dependency>
            <groupId>org.graalvm.regex</groupId>
            <artifactId>regex</artifactId>
            <version>${graalvm.version}</version>
            <exclusions>
                <exclusion>
                    <groupId>org.graalvm.sdk</groupId>
                    <artifactId>graal-sdk</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

lampajr (changes requested):
Hi @willr3 , looks like there is a failure compiling the `horreum-integration-tests` module:
```bash
[ERROR] Rule 0: org.apache.maven.enforcer.rules.dependency.DependencyConvergence failed with message:
[ERROR] Failed while enforcing releasability.
[ERROR] 
[ERROR] Dependency convergence error for org.graalvm.regex:regex:jar:23.0.1 paths to dependency are:
[ERROR] +-io.hyperfoil.tools:horreum-integration-tests:jar:0.13-SNAPSHOT
[ERROR]   +-io.hyperfoil.tools:horreum-backend:jar:0.13-SNAPSHOT:compile
[ERROR]     +-org.graalvm.js:js:jar:23.0.1:compile
[ERROR]       +-org.graalvm.regex:regex:jar:23.0.1:compile
[ERROR] and
[ERROR] +-io.hyperfoil.tools:horreum-integration-tests:jar:0.13-SNAPSHOT
[ERROR]   +-io.hyperfoil.tools:horreum-backend:jar:0.13-SNAPSHOT:compile
[ERROR]     +-org.graalvm.regex:regex:jar:23.0.3:compile
```

I think the easiest solution is the additional exclusions I suggested below.

horreum-backend/pom.xml
willr3: I think this was based on when we used graal `23.0` but `mvn clean install` passes now after the switch to `23.1`. Was this recommendation based on the failing test or something else you saw in the pom?

horreum-backend/pom.xml
willr3: I think this issue was due to the conflicting versions in quarkus and the declared graal version. I think it is not currently needed for the current version of quarkus and graalvm. Please let me know if you still think it is necessary

horreum-backend/pom.xml
lampajr: The change was required, otherwise the `integration-tests` module couldn't compile.
But as you said, this is no longer required given that we upgraded the version.

lampajr (approved):
lgtm

johnaohara (approved):

Issue #1092, opened 2024-01-12
Reporting/Data Integration API


johnaohara:
### Discussed in https://github.com/Hyperfoil/Horreum/discussions/916

<div type='discussions-op-text'>

<sup>Originally posted by **johnaohara** November 17, 2023</sup>
Horreum ingests, stores, calculates derived metrics and monitors for change.  One core feature of Horreum is automated change detection.  

Currently, there a multiple ways of displaying derived data in the UI, and some rudimentary methods for querying derived data in the system.

There are plenty of visualization and analytical tools available to provide anslysis of datasets, and complex visualizations/dashboards.  

Horreum should not try to emulate/replace with these tools, but should integrate well with tools that users are already familiar with

This requires horreum to define an interface to the derived data, such that it is easy to query and integrate into external tools</div>

willr3:
I've started working on the `/test/` api in https://github.com/willr3/Horreum/tree/test-api

My initial work is on `/labelValues`:
* Change the return from `[{name:"x",value:"y"},...]` to `{x:"y"...}` for the values from each dataset to be easier for grafana consumption and match how the values are used in `TestDatasets.tsx`
* Added `runId` and `datasetId` to the response for each dataset so that values can be correlated to runs
* Added a `filter` option that supports either a sub-json object or a jsonpath query for filtering datasets
  Passing in `{'foo':'bar'}` will only include datassets where the `foo` label value is `bar`
  Passing in `$.foo ? (@ > 5 && @ < 10)` would only return labelValues for datasets where `5 < foo < 10`.
* Added `sort`, `limit`, and `page` parameters.

`sort` expects the name of a label (validation pending) but we could change it to support value selection inside a label similar to how filter works.
I'd also like to discuss removing the `metric` and `filtering` attributes from label values. A user could realistically want to filter based on any label value.

TODO for `/labelValues`:

- [ ]  Explore how errors are returned for API call
- [ ]  Fix testing (existing tests of `/labelValue` return type are commended out)
- [ ] Decide on `sort` options (label names or full jsonpath)

stalep:
Are we happy with this feature wise?
Looking at the doc, we should improve/add docs on how to use this API. It is partially documented in the Grafana integration, but there is no documentation on filtering eg. It should provide examples on how to use it also with curl I think?

johnaohara:
I think we need to add the docs , with examples of the format of the queries before closing. Users need to be able to understand what they can pass into the api

[...]

PR #1322, opened 2024-02-13
fix #1321 added sorting & filtering, and start, stop

### LINKED ISSUE #1321: labelValue sorting and filtering based on time or labelValues
the `/api/test/${testId}/labelValues` endpoint does not include sorting by `start` or `stop` and does not properly filer or order results.

willr3:
Closes #1321

More API improvements related to #1092 for downstream tools to select only the desired data from Horreum


stalep:
Looks good other than some empty lines. - but should we update the docs with examples as part of this pr as well?

willr3:
The docs mention the endpoint for grafana and I'm honestly not sold this is the proper implementation for filter and sort. I based the implementation on what I saw for other components so that we could have the functionality to use while we discuss a consistent `filter` and `sort` design. Once the design is consistent then I agree we should include those features in tutorials / examples.

johnaohara:
@willr3 Did the changes to the api not cause updates to the openapi.yaml?

PR #1324, opened 2024-02-14
fix #1323 by adding /api/run/{id}/labelValues

### LINKED ISSUE #1323: add `/run/${runId}/labelValues` similar to `test/${testId}/labelValues`
**Feature idea.**

Related to #1092

The API has a label related endpoint for `test` and `dataset` but not `run`. I have my opinions about the existence of `datasets` but the user should be able to retrieve all `labelValues` from a run without having to filter on the `test` endpoint.

**Describe the solution you'd like**

We should add `/run/${runId}/labelValues` to match the endpoints that exist for `test` and `run`

**Additional information**

Any additional information you can provide, like an overall design description

willr3:
Closes #1323

related to #1092

this adds the `/api/run/{id}/labelValues` endpoint to mirror `/api/test/{id}/labelValues` and it uses a similar format
```
[ { datasetId, values: { [labelName]: labelValue},...]
```

TODO:
- [x] Still needs tests for the new endpoint
- [x] Need to create a reusable record for the return type

### DIFF
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -954,6 +954,75 @@ paths:
             application/json:
               schema:
                 type: string
+  /api/run/{id}/labelValues:
+    get:
+      tags:
+      - Run
+      description: Get all the label values for the run
+      operationId: labelValues
+      parameters:
+      - name: id
+        in: path
+        description: Run Id
+        required: true
+        schema:
+          format: int32
+          type: integer
+        example: 101
+      - name: filter
+        in: query
+        description: either a required json sub-document or path expression
+        schema:
+          default: "{}"
+          type: string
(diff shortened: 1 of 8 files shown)


willr3:
> I think this would be good to have, +1 Please add docs as part of the todo :)

What do you mean by docs? Do you mean the annotations that automatically document the api endpoint or do you want something written for the `docs` folder? I'm adding the annotations so `openapi.yaml` has a description like other endpoints

willr3:
> Need to create a reusable record for the return type
The #1326 PR creates a `LabelValueMap` that can be used for part of the concrete type so this is waiting for that PR

willr3:
I think it is ready for review and merge

willr3:
rebased and squashed after the fix to #1321

PR #1326, opened 2024-02-14
fix 1325 with concrete return types

### LINKED ISSUE #1325: `DatasetSummary.view` is descriped as string in the schema but is a json object in the response
```
"view": {
  "35": {}
}
```

willr3:
openapi.yaml generation from annotations is a dark art I don't yet fully undestand. I would hope the annotations offer the same expression as we can achieve with concrete types or by had writing the yaml but I cannot find the solution.

Instead I changed the data type to a subclass of `HashMap<>` so that openapi will generate a custom schema and reference it for docs and clients.

Closes #1325

### DIFF
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -2531,8 +2531,12 @@ components:
           type: string
           example: Run on AWS with m7g.large
         view:
-          description: View definition
-          type: string
+          description: map of view component ids to the LabelValueMap to render the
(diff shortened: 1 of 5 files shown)


stalep:
The CI fails, try to add a default constructor to IndexedLabelValueMap.

willr3:
> The CI fails, try to add a default constructor to IndexedLabelValueMap.

Yep, looks like it was angry there wasn't an empty constructor. I'll squash after the remote build works too

PR #1335, opened 2024-02-19
python-client: add Schema + impl to ObjectNode api returns

willr3:
This gets us to only 2 python client generation warnings, both about `text/plain` body content in a `post`

```
#> openapi-python-client generate --path /tmp/openapi.yaml
Generating horreum-rest-api-client
Warning(s) encountered while generating. Client was generated, but some pieces may be missing

WARNING parsing POST /api/run/{id}/description within run. Endpoint will not be generated.

Unsupported content type text/plain


WARNING parsing POST /api/run/{id}/schema within run. Endpoint will not be generated.

Unsupported content type text/plain


If you believe this was a mistake or this tool is missing a feature you need, please open an issue at https://github.com/openapi-generators/openapi-python-client/issues/new/choose
```
The issue is already opened with `openapi-python-client` 
https://github.com/openapi-generators/openapi-python-client/discussions/821

### DIFF
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -1967,12 +1967,27 @@ paths:
           type: string
         example:
           key: requiredValue
-      - name: sort
+      - name: before
         in: query
-        description: label name for sorting
+        description: ISO-like date time string or epoch millis
         schema:
           default: ""
           type: string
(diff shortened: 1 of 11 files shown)


johnaohara:
Which issue does this relate to? Does it close an issue, or is it a partial impl for an existing issue?

Issue #1336, opened 2024-02-19
openapi.yaml ProtectedTimeType has incorrect type and format


willr3:
From: https://github.com/Hyperfoil/horreum-client-python/issues/1#issuecomment-1949316739

The api is returning the millis since epoch but claiming it will return a string in `openapi.yaml`

```
          format: date-time
          description: Run Stop timestamp
          type: string
          example: 2019-09-26T07:58:30.996+0200
```
should probably be 
```
          format: int64
          description: Run Start timestamp
          type: integer
          example: 1234567890

```

PR #1337, opened 2024-02-19
fix 1336 with corrected start and stop annotations

willr3:
Should generate a corrected openapi.yaml. We should also look at testing the generated openapi.yaml with all endpoints?

closes: #1336

### DIFF
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/ProtectedTimeType.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/ProtectedTimeType.java
@@ -11,13 +11,13 @@ import java.time.Instant;
 public class ProtectedTimeType extends ProtectedType{
     @NotNull
     @JsonProperty(required = true)
-    @Schema(type = SchemaType.STRING, implementation = Instant.class,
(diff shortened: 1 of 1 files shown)


stalep:
We should have a look at how we format date/time in Horreum. In TestService.listLabelValues we use strings as the type, but we accept either iso-like or epoch. I think the intention was to default to that, but it has not been communicated well.

willr3:
`TestService.listLabelValues` supports different time formats for the `before` and `after` query parameters to make it easier for users to request data from Horreum. The issue this addresses is how we describe the data that Horreum provides.

When Horreum encodes an `Instant` into json it is currently encoded as a `Long` but the annotations told openapi to create an `openapi.yaml` that claims the values will be an ISO style string. The PR changes the annotations to match the data type Horreum is already using.

We can discuss if `Long` or an ISO `String` is more appropriate for integration with other tools but for now this PR is a bug fix for incorrect annotations.

webbnh (approved):
Looks good to me.

johnaohara:
This PR is missing the changes to `openapi.yaml`. Please can we ensure that if we change the openapi annotations, that the corresponding generated `openapi.yaml` is also checked in with the PR otherwise what we publish on the website is out of sync with the actual API

PR #1370, opened 2024-02-22
Add Report deletion task. fixes 974

### LINKED ISSUE #974: Document how to delete a report
Update docs on how to delete reports and report configurations

whitingjr:
## Fixes Issue

fixes #974 
 
## Changes proposed

[x] - add documentation on managing Reports and configurations

- [x] My change requires changes to the documentation.


docs/site/content/en/docs/Tasks/manage-reports/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+Creation of Reports in Horreum is straightforward enough but deletion can be not obvious. A Report in Horreum can be updated or saved as a seperate versions (configurations). A useful procedure when modifying an existing Report that functions correctly.
+
+## Report Deletion
+
+{{% imgproc total-reports-on-Report-list-page Fit "460x528" %}}
+Select Report configurations
+{{% /imgproc %}}
+
+To delete an existing report configuration select the folder icon in the Total reports column on the Reports list view. Each instance of a configuration will have a red coloured button named Delete.
```
johnaohara, line 19: The "report configuration" is the list of reports defined at `http://example.com/reports`  a "report"  is an instance of a report configuration populate with data. This sentence is referring to a "report configuration" instead of a "report"

johnaohara (changes requested):
Users need to be able to delete report configurations as well, which should be documented,  but i don't think we expose that functionality in the UI

docs/site/content/en/docs/Tasks/manage-reports/index.md
whitingjr: This really needs to be in `Core Concepts` as well. I'll update this section.

johnaohara: sounds good

whitingjr: Having worked on Reports for a while and modifying the configuration of the Report and saved the new instance the intuitive train of thought is to call those instances a "Report configuration". The distinguishing factor is the configuration itself.
 To call the "Report" in the list a "Report configuration" as you have explained seems unintuitive. Are you adamant it should be described this way ?

johnaohara: The UI refers to the list a "Report configuration" 

image

and the report instances as reports;

image


If we start calling them something else in the docs, then that will be confusing for users

whitingjr: so it does.

whitingjr:
> Users need to be able to delete report configurations as well, which should be documented, but i don't think we expose that functionality in the UI

I will add an example of creating a request to delete the Report Configuration using the REST api.

PR #1428, opened 2024-03-06
Add a GH Action workflow to check package-lock.json file for the commit.

### LINKED ISSUE #1427: Add prettier format checking on package-lock.json
**Describe the bug**

 Changes proposed by dependabot cause unnecessary formatting changes to package-lock.json.

**To Reproduce**

See this [PR](https://github.com/Hyperfoil/Horreum/pull/1425/files)

whitingjr:
## Fixes Issue

fixes #1427 

## Changes proposed

[x] - Add a workflow that checks the format of package-lock

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,4 +1,4 @@
-name: Format check
+name: Format check - package-lock.json
 
 on: [pull_request]
 
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,17 @@
+name: Format check
+
+on: [pull_request]
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install Prettier
+        run: npm install prettier
+
+      - name: Check formatting
+        run: npx prettier --check "horreum-web/package-lock.json"
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,6 +1,9 @@
 name: Format check - package-lock.json
 
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      - 'horreum-web/package-lock.json'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -0,0 +1,17 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install Prettier
+        run: npm install prettier
+
+      - name: Check formatting
+        run: npx prettier --check "package-lock.json"
```
johnaohara, line 17: This will run on root dir, i think this needs to be 
```suggestion
        run: npx prettier --check "horreum-web/package-lock.json"
```

horreum-web/.prettierrc
```diff
@@ -3,6 +3,7 @@
   "trailingComma": "es5",
   "singleQuote": false,
   "printWidth": 120,
-  "tabWidth": 4,
-  "arrowParens": "avoid"  
+  "tabWidth": 2,
```
johnaohara, line 6: are we using `.prettierrc` for anything else?  will changing the existing configuration break something else

.github/workflows/check.yml
```diff
@@ -0,0 +1,17 @@
+name: Format check
```
johnaohara, line 1:
```suggestion
name: Format check - package-lock.json
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on: [pull_request]
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
```
johnaohara, line 7: i think we only want to run this if package-lock.json has been changed, and not on all PR's

johnaohara (changes requested):

.github/workflows/check.yml
whitingjr: sure good idea.

johnaohara (approved):

PR #1430, opened 2024-03-06
Backport: disable no-squash with backport-squash label

lampajr:
## Changes proposed

1. Made `no-squash` option configurable, by default is set to `true` (i.e., using label `backport`) but if you want to force `no-squash: false` for any reason you can simply add label `backport-squash` rather than `backport` one.
2. Added `concurrency` so that if multiple jobs are triggered targeting the same PR only one will be kept alive

- [x] I have updated the documentation accordingly.

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -35,7 +35,17 @@ docker logout quay.io
 
 Backports are automatically created from PRs against the master branch to the stable branch.
 
-To backport a change, add the `backport` label to the PR that you opened against the master branch.
+To backport a change, add the `backport` (or `backport-squash` but not both) label to the PR that you opened against the master branch.
+
+Which label should I use?
+* `backport`: (default) this uses the `no-squash=true` option so that the tool tries to backport all commits coming
+from the original pull request you are trying to backport.
+> _**Note**_ that in this case the commit SHAs should exist during the backporting, i.e,
+delete the source branch only after the backporting PR got created.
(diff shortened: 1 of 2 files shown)


lampajr:
@johnaohara maybe this is just temporarily needed but I think that having this option enabled could be useful for the future (don't know maybe there will be other use cases where we would like to disable `no-squash`)

docs/RELEASE.md
```diff
@@ -35,7 +35,7 @@ docker logout quay.io
 
 Backports are automatically created from PRs against the master branch to the stable branch.
 
-To backport a change, add the `backport` label to the PR that you opened against the master branch.
+To backport a change, add the `backport` (or `backport-squash` but not both) label to the PR that you opened against the master branch.
```
johnaohara, line 38: @lampajr can you briefly describe the behaviour of each label?

lampajr: Sure, doing it!

lampajr: Added here https://github.com/Hyperfoil/Horreum/pull/1430/files#diff-9c4ae3b5f6675180493e9a1a817baabf4c5314870a3fa67c9c6ab89186046a91R40-R48, let me know if that is clear enough

lampajr: image

johnaohara: LGTM

johnaohara:
thanks

PR #1432, opened 2024-03-06
Load all keycloak users on migration

barreiro:
Fixes: #1441 1441


willr3:
This addressed my migration issue and only leaves out users who are in groups we do not plan to migrate

willr3 (approved):

Issue #907, opened 2023-11-14
"Test label calculation" doesn't work with "All match" (array) extractors


dustinblack:
**Describe the bug**

When using a combination function with an extractor in a label that is set to use `All matches` and therefore return an array, The `Test label calculation` button will return an error on any function that attempts to treat the extractor value as an array, as though the value were instead a scalar.

I'm guessing that whether `First match` or `All matches` is set on an extractor, the test always runs with an assumption of `First match`.

The functions appear to work as expected in other views/reports, just not with the test.

![image](https://github.com/Hyperfoil/Horreum/assets/10656942/ceb1b00e-6505-4a21-b5f1-ffd9c06125bf)
![image](https://github.com/Hyperfoil/Horreum/assets/10656942/1460722e-dd7a-47d4-8bc6-54ed7b8ed5ab)

**To Reproduce**

Set any label extractor to `All matches` and write a combination function that performs an operation on the value as an array, and click the `Test label calculation` button.

**Version**

Current hosted: 0.9.6

**OS**

Fedora 39

**Browser**

Brave 1.60.114

lampajr:
I can confirm what Dustin already said, the issue seems happening just in the `Test label calculation` action and the motivation is actually that the `array` field is ignored in the backed due to mismatch in the naming convention between the psql type `Extractor` and the `Extractor` DTO.

The type is created with field named `isarray`: https://github.com/Hyperfoil/Horreum/blob/e250947f4eff0779e2f637d61f54b794d1f5ef4c/horreum-backend/src/main/resources/db/changeLog.xml#L3551-L3553

but the `Extractor` class defines a `array` field: https://github.com/Hyperfoil/Horreum/blob/e250947f4eff0779e2f637d61f54b794d1f5ef4c/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Extractor.java#L23

Therefore, during the `previewLabel` action the extractor is serialized from the `Extractor` DTO: https://github.com/Hyperfoil/Horreum/blob/e250947f4eff0779e2f637d61f54b794d1f5ef4c/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java#L351

[...]

PR #1461, opened 2024-03-08
Fix test label calculation with all match

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/907

## Changes proposed

Fixes the `Extractor` OpeAPI schema object:
* Changes field from `array` to `isarray` as this must match the PSQL `extractor` type

More details on the root cause are described [here](https://github.com/Hyperfoil/Horreum/issues/907#issuecomment-1984089188).


horreum-integration-tests/src/test/java/io/hyperfoil/tools/horreum/it/HorreumClientIT.java
```diff
@@ -129,6 +130,51 @@ public void testJavascriptExecution() throws InterruptedException {
         Assertions.assertEquals("foobar", preview.value.textValue());
     }
 
+    @org.junit.jupiter.api.Test
+    public void testJavascriptExecutionWithArray() throws InterruptedException {
```
lampajr, line 134: Added new IT to ensure the Javascript invocation does work with array objects, the same test fails if run against `master`.

johnaohara (approved):
LGTM

PR #1470, opened 2024-03-09
Add script to clear out user data.

whitingjr:
A simple script that clears out user data from the database. Leaving the 'dummy' account.

## Fixes Issue

Fixes the issue of having personally recognizable data on developers machines.

## Changes proposed

- [x] - Script added

### DIFF
--- a/infra-legacy/clear-data.sh
+++ b/infra-legacy/clear-data.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# This script clears the database of user account information
+
+set -o pipefail
+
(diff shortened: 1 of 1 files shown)


johnaohara:
@whitingjr what problem does this solve?

whitingjr:
@johnaohara I provided details elsewhere. Not available in the open.
 The issue is to do with handling PII information.

johnaohara:
Handling PII is an operational issue, i.e. how do users handle production data, which is a concern that exists outside the project.  

The intended purpose of this PR is to delete data from a production database, but the workflow of retrieving a prod backup, boot-strapping it with horreum to be able to run the script is not the correct process to handle PII. All PII  should have been cleaned **before** a user can even take a copy of the production DB

I do not think we should add scripts to the project that deletes all users from the database, which is production data.

barreiro:
-1 from me as well. If users need to call this script to remove PII from a running postgres container, probably is too late for that :wink:!!

PR #1485, opened 2024-03-14
fix #1484 remove final pattern check

### LINKED ISSUE #1484: Unable to parse start time

willr3:
fix #1484

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
@@ -649,7 +649,7 @@ public class Util {
          if (str.matches("\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}")){
             str=str+"Z";
          }
-         //ISO_DATE_TIME, do not verify with pattern because multipel patterns are possible
+         //ISO_DATE_TIME, do not verify with pattern because multiple patterns are possible
          try {
             return ZonedDateTime.parse(str, DateTimeFormatter.ISO_DATE_TIME).toInstant();
          } catch (DateTimeParseException e) {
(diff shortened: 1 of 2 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
```diff
@@ -649,13 +649,11 @@ public static Instant toInstant(Object time) {
...
-            try {
-               return ZonedDateTime.parse(str, DateTimeFormatter.ISO_DATE_TIME).toInstant();
-            } catch (DateTimeParseException e) {
-               e.printStackTrace();
-            }
+         //ISO_DATE_TIME, do not verify with pattern because multipel patterns are possible
+         try {
+            return ZonedDateTime.parse(str, DateTimeFormatter.ISO_DATE_TIME).toInstant();
+         } catch (DateTimeParseException e) {
+            e.printStackTrace();
```
johnaohara, line 656: This should be logged with the logger

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
```diff
@@ -649,13 +649,11 @@ public static Instant toInstant(Object time) {
...
             str=str+"Z";
          }
-         //ISO_DATE_TIME
-         if(str.matches("\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:Z|[+\\-]\\d{2}:\\d{2})")){
-            try {
-               return ZonedDateTime.parse(str, DateTimeFormatter.ISO_DATE_TIME).toInstant();
-            } catch (DateTimeParseException e) {
-               e.printStackTrace();
-            }
+         //ISO_DATE_TIME, do not verify with pattern because multipel patterns are possible
```
johnaohara, line 652:
```suggestion
         //ISO_DATE_TIME, do not verify with pattern because multiple patterns are possible
```

johnaohara (changes requested):
We should log the `DateTimeParseException` exception with the defined logger rather than using `printStackTrace()`

PR #1495, opened 2024-03-18
fix 1484 and add toInstant testing

### LINKED ISSUE #1484: Unable to parse start time

willr3:
Adds the changes from the `0.12` branch fix for #1484 along with additional testing and improve error logging
closes #1484


johnaohara:
@willr3 this PR will create a diverging fix/test in master branch and 0.12.x, with tests missing from 0.12.

It would be easier to manage if the patch to 0.12 was ported to master branch and the additional tests added in a separate PR that we can backport to 0.12

Not only would that be cleaner (in terms of keeping the changes in branches in sync), it will massively help me with the realease docs for 0.13

PR #1496, opened 2024-03-18
fix#1484 remove final pattern check

### LINKED ISSUE #1484: Unable to parse start time

willr3:
fix #1484


willr3:
Does that mean I got it right or you gave up? :)

johnaohara:
This is the same change that went in `0.12.x` branch so go merge

PR #1498, opened 2024-03-18
Remove general use of Elasticsearch dev service

barreiro:
Create a new test profile, currently only for `DatasourceTest`.


johnaohara (approved):
LGTM

PR #1525, opened 2024-03-20
Remove vertx dispatch and tiddy up UserServiceImpl

barreiro:
The dispatch of some operations (`vertx.executeBlocking(...)`) cause trouble when calling the `UserServiceImpl` directly, like for testing., because lack of request context. 

Took the opportunity to perform some refactoring as well.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
@@ -1,405 +1,340 @@
 package io.hyperfoil.tools.horreum.svc.user;
 
 import io.hyperfoil.tools.horreum.api.internal.services.UserService;
-import io.hyperfoil.tools.horreum.svc.CountDownFuture;
 import io.hyperfoil.tools.horreum.svc.Roles;
 import io.hyperfoil.tools.horreum.svc.ServiceException;
 import io.quarkus.arc.lookup.LookupIfProperty;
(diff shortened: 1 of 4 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
```diff
@@ -1,405 +1,340 @@
...
+            role.setComposite(true);
+            RoleRepresentation.Composites composites = new RoleRepresentation.Composites();
+            composites.setRealm(compositeRoles);
+            role.setComposites(composites);
+        }
+        try {
+            keycloak.realm(realm).roles().create(role);
+        } catch (ClientErrorException e) {
+            if (e.getResponse().getStatus() == Response.Status.CONFLICT.getStatusCode()) {
+                LOG.warnv("Registration of role {0} failed because already exists", roleName);
```
shivam-sharma7, line 271:
```suggestion
                LOG.warnv("Registration of role {0} failed because its already exists", roleName);
```
@barreiro WDYT?

shivam-sharma7 (changes requested):

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
barreiro: @shivam-sharma7 thanks!

shivam-sharma7 (approved):

PR #1536, opened 2024-03-24
fix 1514 with empty string check

### LINKED ISSUE #1514: Server log WARN : DateTimeParseException
**Describe the bug**

The following server WARN messages are being logged:

```
2024-03-19 08:12:33,735 WARN  [io.hyp.too.hor.svc.Util] (executor-thread-22) failed to convert  to timestamp using : java.time.format.DateTimeParseException: Text '' could not be parsed at index 0
	at java.base/java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:2052)
	at java.base/java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1954)
	at java.base/java.time.ZonedDateTime.parse(ZonedDateTime.java:600)
	at io.hyperfoil.tools.horreum.svc.Util.toInstant(Util.java:654)
	at io.hyperfoil.tools.horreum.svc.TestServiceImpl.listLabelValues(TestServiceImpl.java:574)
	at io.hyperfoil.tools.horreum.svc.TestServiceImpl_Subclass.listLabelValues$$superforward(Unknown Source)
	at io.hyperfoil.tools.horreum.svc.TestServiceImpl_Subclass$$function$$10.apply(Unknown Source)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:73)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:62)
	at io.hyperfoil.tools.horreum.server.RolesInterceptor.intercept(RolesInterceptor.java:53)
	at io.hyperfoil.tools.horreum.server.RolesInterceptor_Bean.intercept(Unknown Source)
	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:30)
	at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:27)
	at io.hyperfoil.tools.horreum.svc.TestServiceImpl_Subclass.listLabelValues(Unknown Source)
	at io.hyperfoil.tools.horreum.api.services.TestService$quarkusrestinvoker$listLabelValues_0a886b7b0505ea93c534bdfb2272905dab70d786.invoke(Unknown Source)
	at org.jboss.resteasy.reactive.server.handlers.InvocationHandler.handle(InvocationHandler.java:29)
	at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:141)
	at org.jboss.resteasy.reactive.common.core.AbstractResteasyReactiveContext.run(AbstractResteasyReactiveContext.java:147)
	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:582)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2513)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1538)
	at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:29)
	at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:29)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:833)

```

willr3:
Check if the string value of the time passed to Util.toInstant is blank before trying to convert the value. This will prevent logging the error message when toInstant should not attempt a conversion. Also changed from warn to debug when logging the failed conversion because failed user conversions are expected rather than something that requires a warning.

closes #1514

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
@@ -634,6 +634,9 @@ public class Util {
          return Instant.ofEpochMilli(((Number) time).longValue());
       } else {
          String str = time.toString().trim();
+         if(str.isBlank()){
+            return null;
+         }
          if(str.matches("\\d+")){
             try {
                return Instant.ofEpochMilli(Long.parseLong((String) time));
@@ -653,7 +656,7 @@ public class Util {
          try {
             return ZonedDateTime.parse(str, DateTimeFormatter.ISO_DATE_TIME).toInstant();
          } catch (DateTimeParseException e) {
-            log.warn("failed to convert "+time+" to timestamp using "+str,e);
+            log.debug("failed to convert "+time+" to timestamp using "+str);
          }
       }
       return null;//nothing matched
--- a/horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
+++ b/horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
@@ -9,11 +9,19 @@ import io.quarkus.test.common.QuarkusTestResource;
 import io.quarkus.test.junit.QuarkusTest;
 import io.quarkus.test.junit.TestProfile;
 import io.quarkus.test.oidc.server.OidcWiremockTestResource;
+import org.jboss.logging.Logger;
+import org.jboss.logmanager.Level;
+import org.jboss.logmanager.LogContext;
+import org.jboss.logmanager.formatters.PatternFormatter;
+import org.jboss.logmanager.handlers.OutputStreamHandler;
 import org.junit.jupiter.api.Test;
+
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.time.Instant;
 
-import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.*;
 
 @QuarkusTest
 @QuarkusTestResource(PostgresResource.class)
@@ -21,7 +29,38 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 @TestProfile(HorreumTestProfile.class)
 public class UtilTest {
 
+    public static class StringHandler extends OutputStreamHandler{
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
+        public StringHandler(){
+            setOutputStream(baos);
+            this.setFormatter(new PatternFormatter("%m"));
+            this.setLevel(Level.ALL);
+            this.setAutoFlush(true);
+            LogContext.getLogContext().getLogger("").addHandler(this);
+        }
+
+        public String getLog(){return baos.toString();}
+
+        @Override
+        public void close() throws SecurityException {
+            LogContext.getLogContext().getLogger("").removeHandler(this);
+            super.close();
+            try {
+                baos.close();
+            } catch (IOException e) { /* meh */}
+        }
+    }
+
+    @Test
+    public void toInstant_nulls() throws UnsupportedEncodingException {
+        try(StringHandler handler = new StringHandler()){
+            assertNull(Util.toInstant(null),"null input should return null");
+            assertNull(Util.toInstant(""),"empty string should return null");
+            assertNull(Util.toInstant(" "),"blank string should return null");
+            assertFalse(handler.getLog().contains("DateTimeParseException"),handler.getLog());
+        }
+    }
     @Test
     public void toInstant_valid() throws IOException {
         //already an instant
@@ -38,7 +77,6 @@ public class UtilTest {
         //json
         assertNotNull(Util.toInstant(new TextNode("2020-01-01")),"failed to parse json text YYYY-MM-DD");
         assertNotNull(Util.toInstant(new LongNode(System.currentTimeMillis())),"failed to parse current millis as json node");
-
     }
 
 }


horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
```diff
@@ -9,19 +9,58 @@
...
+        }
+    }
 
+    @Test
+    public void toInstant_nulls() throws UnsupportedEncodingException {
+        try(StringHandler handler = new StringHandler()){
+            assertNull(Util.toInstant(null),"null input should return null");
+            assertNull(Util.toInstant(""),"empty string should return null");
+            assertNull(Util.toInstant(" "),"blank string should return null");
+            assertFalse(handler.getLog().contains("DateTimeParseException"),handler.getLog());
```
lampajr, line 61: Is this `,handler.getLog()` as assert message wanted or just a leftover?

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
```diff
@@ -9,19 +9,58 @@
...
+            LogContext.getLogContext().getLogger("").removeHandler(this);
+            super.close();
+            try {
+                baos.close();
+            } catch (IOException e) { /* meh */}
+        }
+    }
 
+    @Test
+    public void toInstant_nulls() throws UnsupportedEncodingException {
```
lampajr, line 56: I am not a fan of mixed CamelCase and snake_case but that is just a minor nitpick :smile: , then up to you.
 
```suggestion
    public void toInstantNulls() throws UnsupportedEncodingException {
```

The same comment for `toInstant_valid` even if outside of this PR.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
```diff
@@ -653,7 +656,7 @@ public static Instant toInstant(Object time) {
          try {
             return ZonedDateTime.parse(str, DateTimeFormatter.ISO_DATE_TIME).toInstant();
          } catch (DateTimeParseException e) {
-            log.warn("failed to convert "+time+" to timestamp using "+str,e);
+            log.debug("failed to convert "+time+" to timestamp using "+str);
```
lampajr, line 659: > Also changed from warn to debug when logging the failed conversion because failed user conversions are expected rather than something that requires a warning.

I agree

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
```diff
@@ -634,6 +634,9 @@ public static Instant toInstant(Object time) {
          return Instant.ofEpochMilli(((Number) time).longValue());
       } else {
          String str = time.toString().trim();
+         if(str.isBlank()){
+            return null;
+         }
```
lampajr, line 639: Do we want to log something here? because that's actually a similar case of `conversion failure`.
Maybe something like `failed to convert null instant to timestamp`, this could be useful foir debugging purposes if not expected.

lampajr (commented):
Overall looks good to me, added just a couple of comments

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
willr3: I originally thought any invalid input to the function should be logged (thus the issue being opened) but apparently Horreum has a use case where start and stop are not specified and they default to the currrent system timestamp. In that use case the input to `toInstant` results in a blank (or empty) string string.
To me that means it is an expected input case and should not be logged as a problem. 

If we decide to introduce debug or trace logging for user input then perhaps we could add additional logging then but I don't think we need it for the purpose of properly handling input to `toInstant`.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
willr3: I think invalid user input should not be logged by default because I don't think the server logs should be where we want to discover that input. I can certainly see a valid use case to being able to turn on debug to identify invalid user input but users should be able to get a response about their timestamps being invalid without checking the server logs.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
lampajr: yeah I missed that use case, then I agree with your approach!

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
willr3: I like `<methodName>_<test_description>` which I think makes it easier to search for the method name as well as easier to read the conditions in the test. 
I find snake_case easier to read for a description but Java methods are usually camelCase, thus the mixed case. There is also the `test<MethodName><TestDescription>` convention but that gets very crowded AndHarderToReadForMyOldEyes. :)

I agree that `src/main` code should adhere to one casing, `cameCase` being the norm for Java, but I think test function names serve to describe what is being tested not to be used as an API.

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
willr3: The assert ensures none of those cases logged the exception for parsing becasue they are all valid uses of the function. It also adds an example for checking the logged output which is something we probably need to cover in other test conditions (where logging that something happens is necessary).

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
lampajr: I never used `<methodName>_<test_description>` approach but I see a value in that, maybe we could think about adding some sort of shared/global formatting and/or checks to ensure we all follow the same convention in future improvements

lampajr (approved):
Thanks @willr3, lgtm!

willr3:
> Thanks @willr3, lgtm!

If it looks good would you please merge? I think we discussed no self merging so that when we review we have the added motivation of being the one to merge :)

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
willr3: https://github.com/Hyperfoil/Horreum/discussions/1538

PR #1539, opened 2024-03-26
UserService tests

barreiro:
(Also includes some changes in the service and back-ends to make them behave as same as possible) 

Add abstract class with test cases and two concrete implementations, one for each back-end. Each implementation has it's own custom `QuarkusTestProfile`.

The test injects the service and call it's methods directly. This is simpler and allows greater control and therefore better coverage.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/UserServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/UserServiceImpl.java
@@ -13,10 +13,12 @@ import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
 import org.jboss.logging.Logger;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import static java.text.MessageFormat.format;
+import static java.util.Collections.emptyList;
 
 @Authenticated
 @ApplicationScoped
@@ -43,13 +45,8 @@ public class UserServiceImpl implements UserService {
     // ideally we want to enforce these roles in some of the endpoints, but for now this has to be done in the code
     // @RolesAllowed({ Roles.ADMIN, Roles.MANAGER })
     @Override public void createUser(NewUser user) {
-        if (user == null) {
-            throw ServiceException.badRequest("Missing user as the request body");
-        } else if (user.team != null && !user.team.endsWith("-team")) {
-            throw ServiceException.badRequest(format("Team {0} does not end with ´´-team´´ suffix", user.team));
-        }
+        validateNewUser(user);
         userIsManagerForTeam(user.team);
-
         backend.get().createUser(user);
         LOG.infov("{0} created user {1} {2} with username {3} on team {4}", identity.getPrincipal().getName(), user.user.firstName, user.user.lastName, user.user.username, user.team);
     }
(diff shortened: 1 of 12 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/UserServiceImpl.java
```diff
@@ -58,29 +55,41 @@ public class UserServiceImpl implements UserService {
...
     @Override public String defaultTeam() {
         UserInfo userInfo = UserInfo.findById(identity.getPrincipal().getName());
         return userInfo != null ? userInfo.defaultTeam : null;
     }
 
     @Transactional
     @WithRoles(addUsername = true)
-    @Override public void setDefaultTeam(String team) {
+    @Override public void setDefaultTeam(String unsafeTeam) {
+        String team = unsafeTeam != null ? validateTeamName(unsafeTeam) : null;
```
stalep, line 68: So this can set team to null, which will set userInfo.defaultTeam = null, is that intended?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/UserServiceImpl.java
```diff
@@ -58,29 +55,41 @@ public class UserServiceImpl implements UserService {
         return backend.get().getTeams();
     }
 
-    @WithRoles(addUsername = true)
+    @Transactional
+    @WithRoles(extras = Roles.HORREUM_SYSTEM)
     @Override public String defaultTeam() {
         UserInfo userInfo = UserInfo.findById(identity.getPrincipal().getName());
         return userInfo != null ? userInfo.defaultTeam : null;
```
stalep, line 62: Is it "ok" to return null here?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/UserServiceImpl.java
stalep: Would it not be better to do: String name = validateTeamName(unsafeTeam); ?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/DatabaseUserBackend.java
```diff
@@ -127,34 +139,44 @@ private void addTeamMembership(UserInfo userInfo, String teamName, TeamRole role
...
-        String prefix = getTeamPrefix(team);
-        Team teamEntity = Team.find("teamName", prefix.substring(0, prefix.length() - 1)).firstResult();
+        Team teamEntity = Team.find("teamName", removeTeamSuffix(team)).firstResult();
         if (teamEntity == null) {
-            throw ServiceException.notFound("The team " + team + " does not exist");
+            throw ServiceException.notFound(format("The team {0} does not exist", team));
         }
-        TeamMembership.delete("team", teamEntity);
+        // need to delete memberships with roles
+        deleteTeamAndMemberships(teamEntity);
```
stalep, line 162: It this throws an exception it would be propagated to the client directly if I'm not mistaken. If so we should add some logic to capture that.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/DatabaseUserBackend.java
```diff
@@ -127,34 +139,44 @@ private void addTeamMembership(UserInfo userInfo, String teamName, TeamRole role
...
+
+    @Transactional
+    @WithRoles(fromParams = DeleteTeamAndMembershipsParameterConverter.class)
+    void deleteTeamAndMemberships(Team teamEntity) {
+        // need to remove from the "owning" side of the relationship
+        teamEntity.teams.stream().map(membership -> membership.user.username).forEach(username -> {
+            Optional<UserInfo> user = UserInfo.findByIdOptional(username);
+            user.ifPresent(u -> u.teams.removeIf(membership -> teamEntity.equals(membership.team)));
+        });
         teamEntity.delete();
```
stalep, line 173: Similar as above

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/UserServiceImpl.java
barreiro: both approaches are reasonable. it depends if we want to allow `defautTeam` to be unset or not.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/UserServiceImpl.java
barreiro: probably safer to return `""` when `userInfo` exists.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/UserServiceImpl.java
barreiro: changed to not allow unsettling `defaultTeam`

PR #1549, opened 2024-03-28
Remove hypersistence dependency

barreiro:
This dependency will be problematic for a Quarkus upgrade.

It looks to me that it can be safely removed.

### DIFF
--- a/horreum-backend/pom.xml
+++ b/horreum-backend/pom.xml
@@ -41,11 +41,6 @@
             </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


stalep:
I know we had some issues when we looked at removing it in the past. Did you test this against a prod db and checked if it worked fine?

barreiro:
I relied mostly on the test suite, but will do check with the prod backup as well and see if I can spot any issues.

barreiro:
I did click around with production data and did not encounter any issues.

PR #1558, opened 2024-03-30
Add type safety to sequence generators

barreiro:


stalep:
> > looks good, but with this change I thing we also should delete `io.hyperfoil.tools.horreum.entity.SeqIdGenerator` ?
> 
> @stalep have we separated all the update/add logic from entities. I think this is needed if we still update/add entities with an `em.merge`?

Good point, we have not...

johnaohara (approved):

johnaohara:
I think we will cover removing `io.hyperfoil.tools.horreum.entity.SeqIdGenerator` in a seperate PR, as this would mean changes to the API and is more intrusive

PR #1568, opened 2024-04-03
increase txn timeout to 120s

willr3:


willr3:
Created to formalize the timeout change but discussing the correct location to merge the configuration option

johnaohara:
I think we should handle this in the prod configuration and work to fix the root cause of the slow query that currently needs this setting once the run table gets to a certain size : https://github.com/Hyperfoil/Horreum/issues/1580

Issue #1579, opened 2024-04-09
GH-1577 / Make processNewOrUpdatedSchema asynchronous


lampajr:
Short term fix for of https://github.com/Hyperfoil/Horreum/issues/1577

**Feature idea.**

Since fixing https://github.com/Hyperfoil/Horreum/issues/1577 in the proper way could require some additional refactoring, that could have other implication, as short-term goal we decided to make the `processNewOrUpdatedSchema` https://github.com/Hyperfoil/Horreum/blob/3849f043bb457deb86df5fb8ecab5fc8fc3de588/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java#L176-L182
asynchronous, such that the creation/update of a `Schema` would become quite immediate and the `run_schemas` update (especially the `FIND_RUNS_WITH_URI` query execution) is handled asynchronously.

**Describe the solution you'd like**

Create a new channel to manage this new event type, on event receiving, trigger the `processNewOrUpdatedSchema` method.

**Additional information**

n/a

lampajr:
Worth to note that with this (temporary) approach we should stick with an increased transaction timeout (~120s) otherwise the `FIND_RUNS_WITH_URI` query will go in timeout even in the async process.

willr3:
This approach should also include some update to the UI so that the user knows the change is being processed in the background. There are parts of UI (e.g. assigning a run's schema) that relied on this being a synchronous operation.

lampajr:
> This approach should also include some update to the UI so that the user knows the change is being processed in the background. There are parts of UI (e.g. assigning a run's schema) that relied on this being a synchronous operation.

Good point, since this is meant to be (hopefully) a temporary solution, do you think it would be enough to show a popup/warning/alert on schema save?

willr3:
Something in the banner display or any other inline popup might work. I would avoid any use of `alert` or a modal.

Issue #1438, opened 2024-03-07
Create configuration guide for Administrators


johnaohara:
**Describe the bug**

Produce a configuration guide for Horreum system administrators which documents system-wide configuration for setting up and running a Horreum instance

e.g. document features such as: https://github.com/Hyperfoil/Horreum/pull/1437

PR #1581, opened 2024-04-09
MPL-68: make run_schemas updates asynchronous

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/1579

## Changes proposed

Changes:
- [x] create new channel named `schema-sync-{in, out}` and related event `CreateOrUpdateEvent` (TBH I think this could be simplified further as we just need the `schemaId` value, thus we can remove the event and keep an id only)
- [x] move `newOrUpdatedSchema` from `ServiceMediator` to `SchemaServiceImpl` for coherence (especially as we need the TxManager)
- [x] make `runService.onNewOrUpdatedSchema(event.id);` asynchronous

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ServiceMediator.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ServiceMediator.java
@@ -10,7 +10,6 @@ import io.hyperfoil.tools.horreum.api.data.TestExport;
 import io.hyperfoil.tools.horreum.api.services.ExperimentService;
 import io.hyperfoil.tools.horreum.bus.AsyncEventChannels;
 import io.hyperfoil.tools.horreum.entity.data.ActionDAO;
-import io.hyperfoil.tools.horreum.entity.data.SchemaDAO;
 import io.hyperfoil.tools.horreum.events.DatasetChanges;
 import io.smallrye.reactive.messaging.annotations.Blocking;
 import io.vertx.core.Vertx;
@@ -84,6 +83,10 @@ public class ServiceMediator {
     @OnOverflow(value = OnOverflow.Strategy.BUFFER, bufferSize = 10000)
     @Channel("run-recalc-out")
     Emitter<Integer> runEmitter;
+
+    @OnOverflow(value = OnOverflow.Strategy.BUFFER, bufferSize = 10000)
+    @Channel("schema-sync-out")
+    Emitter<Integer> schemaEmitter;
 
     private Map<AsyncEventChannels, Map<Integer, BlockingQueue<Object>>> events =  new ConcurrentHashMap<>();
 
@@ -168,6 +171,18 @@ public class ServiceMediator {
         runEmitter.send(runId);
     }
 
+    @Incoming("schema-sync-in")
+    @Blocking(ordered = false, value = "horreum.schema.pool")
+    @ActivateRequestContext
+    public void processSchemaSync(int schemaId) {
+        runService.onNewOrUpdatedSchema(schemaId);
+    }
+
+    @Transactional(Transactional.TxType.NOT_SUPPORTED)
+    void queueSchemaSync(int schemaId) {
+        schemaEmitter.send(schemaId);
+    }
+
     void dataPointsProcessed(DataPoint.DatasetProcessedEvent event) {
         experimentService.onDatapointsCreated(event);
     }
@@ -216,9 +231,6 @@ public class ServiceMediator {
             subscriptionService.importSubscriptions(test);
     }
 
-    public void newOrUpdatedSchema(SchemaDAO schema) {
-        runService.processNewOrUpdatedSchema(schema);
-    }
     public void updateFingerprints(int testId) {
         datasetService.updateFingerprints(testId);
     }
@@ -246,7 +258,7 @@ public class ServiceMediator {
     }
 
     public <T> BlockingQueue<T> getEventQueue(AsyncEventChannels channel, Integer id) {
-        if (testMode ) {
+        if (testMode) {
             events.putIfAbsent(channel, new HashMap<>());
             BlockingQueue<?> queue = events.get(channel).computeIfAbsent(id, k -> new LinkedBlockingQueue<>());
             return (BlockingQueue<T>) queue;
--- a/horreum-backend/src/main/resources/application.properties
+++ b/horreum-backend/src/main/resources/application.properties
@@ -46,6 +46,18 @@ mp.messaging.outgoing.run-recalc-out.address=run-recalc
 mp.messaging.outgoing.run-recalc-out.durable=true
 mp.messaging.outgoing.run-recalc-out.container-id=horreum-broker
 mp.messaging.outgoing.run-recalc-out.link-name=run-recalc
+# schema-sync incoming
+mp.messaging.incoming.schema-sync-in.connector=smallrye-amqp
+mp.messaging.incoming.schema-sync-in.address=schema-sync
+mp.messaging.incoming.schema-sync-in.durable=true
+mp.messaging.incoming.schema-sync-in.container-id=horreum-broker
+mp.messaging.incoming.schema-sync-in.link-name=schema-sync
+# schema-sync outgoing
+mp.messaging.outgoing.schema-sync-out.connector=smallrye-amqp
+mp.messaging.outgoing.schema-sync-out.address=schema-sync
+mp.messaging.outgoing.schema-sync-out.durable=true
+mp.messaging.outgoing.schema-sync-out.container-id=horreum-broker
+mp.messaging.outgoing.schema-sync-out.link-name=schema-sync
 
 ## Datasource updated by Liquibase - the same as app but always with superuser credentials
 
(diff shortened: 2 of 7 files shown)


horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceTest.java
lampajr, line 1: Added a couple of tests, not sure if this is enough but at least it is a start.

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Schema.java
```diff
@@ -74,4 +74,23 @@ public String toString() {
...
+            this.uri = uri;
+        }
+        @Override
+        public String toString() {
+            return "CreateOrUpdateEvent{" +
+                    "id=" + id +
+                    ", uri=" + uri +
+                    '}';
+        }
+    }
```
johnaohara, line 95: I don't think we want to expose this class in the API, it is an internal type

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ServiceMediator.java
```diff
@@ -168,6 +167,18 @@ void queueRunRecalculation(int runId) {
         runEmitter.send(runId);
     }
 
+    @Incoming("schema-sync-in")
+    @Blocking(ordered = false, value = "horreum.run.pool")
```
johnaohara, line 171: We will want to configure this pool size independent of the `horreum.run.pool`, We should introduce a new configuration option to size this consumer pool

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceTest.java
johnaohara: Do the tests at least cover the use cases that caused the issue?

johnaohara (changes requested):
There are a couple of areas that I think we need to resolve.

This change also requires changes to production configuration. Ideally we would want an update to an "admin guide" (that we don't have: https://github.com/Hyperfoil/Horreum/issues/1438).   I don't think admin is a blocker, but have linked here to reference when we do create an admin guide

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ServiceMediator.java
lampajr: Yeah, you are right.

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Schema.java
lampajr: Yeah, makes sense.. I think I will simplify the event further (it should work with just an Integer as highlighted in the pr description). This way it won't be exposed anywhere as no more existing.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ServiceMediator.java
lampajr: Added `horreum.schema.pool` configuration with commit https://github.com/Hyperfoil/Horreum/pull/1581/commits/2a5aa2fa27d280b24018c9787b4a8ad27b2737de

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Schema.java
lampajr: Completely removed with commithttps://github.com/Hyperfoil/Horreum/commit/2a5aa2fa27d280b24018c9787b4a8ad27b2737de

horreum-backend/src/main/resources/application.properties
```diff
@@ -76,6 +88,7 @@ horreum.test-mode=false
 # thread pool sizes
 smallrye.messaging.worker.horreum.dataset.pool.max-concurrency=10
 smallrye.messaging.worker.horreum.run.pool.max-concurrency=10
+smallrye.messaging.worker.horreum.schema.pool.max-concurrency=10
```
willr3, line 91: This increases the total number of `max-concurrency` to 30 but `quarkus.datasource.jdbc.max-size=25`
Can this lead to connection starvation for the UI?

lampajr: Do you see an issue in increasing the number of `quarkus.datasource.jdbc.max-size`? Otherwise I think we can reduce the `smallrye.messaging.worker.horreum.schema.pool.max-concurrency` as I am not expecting many schemas changes happen in parallel

stalep: I know Jesper said our db could handle 100 connections, but now we're not running it on a separate machine any longer. I would be temped to keep the maximum number of connections and reduce each worker pool to 7 or something?

johnaohara: We need to pick sensible defaults to run OOTB on any machine, the **actual** configuration should be defined at runtime to match the size of the machine / environment it is running on. The default configuration should not be configured for our (current) production environment.

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/Schema.java
johnaohara: thanks

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceTest.java
```diff
@@ -273,12 +273,84 @@ public void testFindUsages(TestInfo info) throws InterruptedException  {
...
-                .then().statusCode(200).extract().body().as(List.class);
+      List<SchemaService.LabelLocation> usages =
+              jsonRequest().get("/api/schema/findUsages?label=".concat("category"))
+              .then().statusCode(200).extract().body().as(List.class);
 
       assertNotNull(usages);
+   }
+
+   @org.junit.jupiter.api.Test
+   public void testCreateSchemaAfterRun() throws InterruptedException {
```
willr3, line 284: Should we mention that the run is an array run in the test name? The current Horreum implementation handles arrays and json objects in a significantly different manner.
Maybe `testCreateSchemaAfterRunUsingArrayData`?

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceTest.java
```diff
@@ -273,12 +273,84 @@ public void testFindUsages(TestInfo info) throws InterruptedException  {
...
+      schema.uri = "urn:new-dummy:schema";
+      Schema updatedSchema = addOrUpdateSchema(schema);
+      assertNotNull(updatedSchema);
+      assertEquals(schema.id, updatedSchema.id);
+
+      TestUtil.eventually(() -> {
+         Util.withTx(tm, () -> {
+            em.clear();
+            List<?>  runSchemas = em.createNativeQuery("SELECT * FROM run_schemas WHERE runid = ?1").setParameter(1, runId).getResultList();
+            // two records as the run is an array of two objects, both referencing the same schema
```
willr3, line 350: This comment escaped the copy, paste, edit :)

horreum-web/src/Banner.tsx
```diff
@@ -3,19 +3,58 @@ import { useEffect, useState } from "react"
...
 export default function Banner() {
     const [banner, setBanner] = useState<BannerData>()
     const [updateCounter, setUpdateCounter] = useState(0)
     useEffect(() => {
-        setTimeout(() => setUpdateCounter(updateCounter + 1), 60000)
+        const timeoutId = setTimeout(() => setUpdateCounter(updateCounter + 1), 60000)
         bannerApi.get().then(setBanner)
+
+        return () => clearTimeout(timeoutId)
     }, [updateCounter])
```
willr3, line 54: I know this isn't a line you changed but I believe we need to include `setUpdateCounter` in the dependency array for this `useEffect` because it is entirely possible for React to regenerate that function and our code to left with a pointer to a function that no longer actually sets the update counter.

willr3 (commented):
I picked up on mostly minor changes. If the logic works for array run data I assume it will work for object run data but perhaps we should create an explicit test to ensure that also works as expected?

horreum-web/src/Banner.tsx
lampajr: I am not that sure about that as the `setUpdateCounter` is generated by React's `useState` which should be stable.

I found this [comment](https://stackoverflow.com/questions/71942624/should-a-setstate-function-be-a-dependency-of-useeffect-when-passed-via-hook#:~:text=You%20know%20that%20(because%20the%20setter%20is%20guaranteed%20to%20be%20stable%20by%20useState%2C%20and%20you%27re%20passing%20it%20verbatim%20through%20context%20and%20your%20useMyState%20hook)) on Stackoverflow mentioning it.

horreum-backend/src/main/resources/application.properties
lampajr: As default configuration here I kept 7,7,7

lampajr:
With latest commit https://github.com/Hyperfoil/Horreum/pull/1581/commits/5e9e17a73e9f2478df0eb68056c7f9e44f41fef4 I added a couple of tests covering other slightly different use cases as suggested https://github.com/Hyperfoil/Horreum/pull/1581#pullrequestreview-1991759034.

If you guys agree on this and this is ready to be merged, I will squash all commits before merging.

johnaohara:
@willr3 please can you review this PR again after the update?

willr3 (commented):
Looks good to me

lampajr:
> Looks good to me

Thanks Will, I am going to squash all commits then

johnaohara (approved):
LGTM

willr3 (commented):
lgtm

PR #1591, opened 2024-04-14
Remove unused object from vite.config.ts

shivam-sharma7:
We are not using this from long time and it is done by quarkus so Removing this.

 

### DIFF
--- a/horreum-web/vite.config.ts
+++ b/horreum-web/vite.config.ts
@@ -7,11 +7,8 @@ export default defineConfig({
     base: '/',
     plugins: [react(), viteTsconfigPaths()],
     server: {    
-        // this ensures that the browser DOES NOT open upon server start
-        open: false,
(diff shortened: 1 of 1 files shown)


shivam-sharma7:
> This was added because vite was automatically opening a browser tab to the node dev url. Are you saying vite no longer automatically opens a browser tab by default?

No, I mean once our application start then we can access `3000` port and we don't need to do `npm run dev` to open this.

willr3:
I think your PR is trying to remove a change that was intentionally introduced to prevent view from opening a browser tab.

shivam-sharma7:
> I think your PR is trying to remove a change that was intentionally introduced to prevent vite from opening a browser tab.

Yeah, for sure and I think we don't need it.

johnaohara:
the `open` property appears defaults to false, removing this property retains the current behaviour

PR #1612, opened 2024-04-18
Fix SecurityMigration realm

barreiro:
the user management should rely only on the configuration of the `keycloak-admin` extension. that is independent of the OICD configuration.


johnaohara:
@barreiro there are failing tests.

please can you also add a  description to this PR about the intention of the change

PR #1643, opened 2024-04-30
remove cast to String for JSON datatypes - Fixes #1642

### LINKED ISSUE #1642: Large Allocations of mapping jsonb data in queries
**Describe the bug**

All jsonb data mapping passes through `io.hyperfoil.tools.horreum.hibernate.JsonBinaryType`, which stores the json value to a string before passing into the Jackson ObjecMapper to convert to a JSON type.

This creates a lot of ephemeral objects on the heap, and adds to GC pressure, e.g. for querying `api/test/317/labelValues`, the majority of allocations on heap are caused by:

image

Which leads to 28.4% of CPU time in GC activity:

image

Removing the conversion to string reduces the GC CPU time to 5.32%

image

johnaohara:
## Fixes Issue
Fixes #1642


lampajr (approved):
lgtm

PR #1655, opened 2024-05-02
Move JPA auth to new persistence unit

barreiro:


horreum-backend/src/main/resources/application.properties
```diff
@@ -93,6 +93,13 @@ smallrye.messaging.worker.horreum.schema.pool.max-concurrency=7
 
 hibernate.jdbc.time_zone=UTC
 
+# security JPA
+quarkus.security-jpa.persistence-unit-name=user-auth
+
+quarkus.hibernate-orm."user-auth".datasource=migration
```
barreiro, line 99: I've re-used `migration` datasource here for simplicity, could create a dedicated one instead.

stalep:
What issue does this fix @barreiro ?

barreiro:
superseded by #1660

PR #1656, opened 2024-05-03
Setup unit testing bypassing REST

lampajr:
## Changes proposed

With this PR I am proposing teh following improvements:
- [x] Setup new tests bypassing the REST, this way we can directly inject and call the services we want to test (I added them for `SchemaService` and `TestService`).
- [x] Improve test coverage for `TestService`, this won't be a huge improvement but at least a starting point
   * Before:
TestServiceCoverageBefore
   * After:  
TestServiceCoverageAfter

- [x] Make some fixes or improvements that I noticed while implementing additional tests.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
@@ -71,7 +71,6 @@ public class SchemaServiceImpl implements SchemaService {
    private static final Logger log = Logger.getLogger(SchemaServiceImpl.class);
 
    //@formatter:off
-   private static final String UPDATE_TOKEN = "UPDATE schema SET token = ? WHERE id = ?";
    private static final String FETCH_SCHEMAS_RECURSIVE =
          """
          WITH RECURSIVE refs(uri) AS
@@ -94,7 +93,7 @@ public class SchemaServiceImpl implements SchemaService {
          .addMetaSchema(JsonMetaSchema.getV6())
          .addMetaSchema(JsonMetaSchema.getV7())
          .addMetaSchema(JsonMetaSchema.getV201909()).build();
-   private static final String[] ALL_URNS = new String[] {"urn", "uri", "http", "https", "ftp", "file", "jar"};
+   private static final String[] ALL_URNS = new String[]{"urn", "uri", "http", "https", "ftp", "file", "jar"};
 
    @Inject
    EntityManager em;
@@ -123,7 +122,7 @@ public class SchemaServiceImpl implements SchemaService {
    @WithRoles
    @PermitAll
    @Override
-   public Schema getSchema(int id, String token){
+   public Schema getSchema(int id, String token) {
       SchemaDAO schema = SchemaDAO.find("id", id).firstResult();
       if (schema == null) {
          throw ServiceException.notFound("Schema not found");
@@ -144,7 +143,7 @@ public class SchemaServiceImpl implements SchemaService {
    @WithRoles
    @Transactional
    @Override
-   public Integer add(Schema schemaDTO){
+   public Integer add(Schema schemaDTO) {
       verifyNewSchema(schemaDTO);
       // Note: isEmpty is true for all non-object and non-array nodes
       if (schemaDTO.schema != null && schemaDTO.schema.isEmpty()) {
@@ -153,23 +152,22 @@ public class SchemaServiceImpl implements SchemaService {
       SchemaDAO schema = SchemaMapper.to(schemaDTO);
       if (schemaDTO.id != null && schemaDTO.id > 0) {
          SchemaDAO existing = SchemaDAO.findById(schema.id);
-         if(existing == null)
+         if (existing == null)
             throw ServiceException.badRequest("An id was given, but it does not exist.");
          em.merge(schema);
          em.flush();
-         if(!Objects.equals(schema.uri, existing.uri) ||
-                 Objects.equals(schema.schema, existing.schema)) {
+         if (!Objects.equals(schema.uri, existing.uri) ||
+               Objects.equals(schema.schema, existing.schema)) {
             //We need to delete from run_schemas and dataset_schemas as they will be recreated
             //when we create new datasets psql will still create new entries in dataset_schemas
             // https://github.com/Hyperfoil/Horreum/blob/master/horreum-backend/src/main/resources/db/changeLog.xml#L2522
             em.createNativeQuery("DELETE FROM run_schemas WHERE schemaid = ?1")
-                    .setParameter(1, schema.id).executeUpdate();
+                  .setParameter(1, schema.id).executeUpdate();
             em.createNativeQuery("DELETE FROM dataset_schemas WHERE schema_id = ?1")
-                    .setParameter(1, schema.id).executeUpdate();
+                  .setParameter(1, schema.id).executeUpdate();
             newOrUpdatedSchema(schema);
          }
-      }
-      else {
+      } else {
          schema.id = null;
          schema.persist();
          em.flush();
@@ -202,7 +200,7 @@ public class SchemaServiceImpl implements SchemaService {
    @WithRoles
    @Override
    public SchemaQueryResult list(String roles, Integer limit, Integer page, String sort,
-                                 @DefaultValue("Ascending")  SortDirection direction) {
+                                 @DefaultValue("Ascending") SortDirection direction) {
       PanacheQuery<SchemaDAO> query;
       Set<String> actualRoles = null;
       if (Roles.hasRolesParam(roles)) {
@@ -231,7 +229,7 @@ public class SchemaServiceImpl implements SchemaService {
          query.page(Page.of(page, limit));
       }
 
-      return new SchemaQueryResult( query.list().stream().map(SchemaMapper::from).toList(), SchemaDAO.count());
+      return new SchemaQueryResult(query.list().stream().map(SchemaMapper::from).toList(), SchemaDAO.count());
    }
 
    @WithRoles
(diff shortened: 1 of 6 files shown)


lampajr:
Unfortunately, I noticed only right now that I introduced some formatting changes :disappointed:

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
```diff
@@ -683,16 +688,17 @@ public Integer addOrUpdateLabel(int schemaId, Label labelDTO) {
...
          emitLabelChanged(label.id, schemaId);
       } else {
-         LabelDAO existing = LabelDAO.findById(label.id);
-         if (existing == null) {
-            label.id = -1;
-            existing = label;
-         }
+         LabelDAO existing = (LabelDAO) LabelDAO.findByIdOptional(label.id)
+               .orElseThrow(() -> ServiceException.notFound("Label " + label.id + " not found"));
+
```
johnaohara, line 701: This changes behaviour slightly, before if "existing" was not found, it was set to label, and label.idset to -1

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
```diff
@@ -272,13 +270,19 @@ public void dropToken(int id) {
    }
 
    public String updateToken(int id, String token) {
-      Query query = em.createNativeQuery(UPDATE_TOKEN);
```
johnaohara, line 275: I am +100 for removing native queries, we lose type safety in the code and couple ourselves to the db schema with every native query we have, esp for simple cases like this

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -157,9 +154,10 @@ public Test getByNameOrId(String input){
       TestDAO test = null;
       if (input.matches("-?\\d+")) {
          int id = Integer.parseInt(input);
+         // there could be some issue if name is numeric and corresponds to another test id
```
johnaohara, line 157: Can you open a new issue for this?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -471,37 +476,36 @@ public void dropToken(int testId, int tokenId) {
...
-         throw ServiceException.forbidden("Changing access of Test " + id + " is not permitted. You are not defined as a `tester` for the `"+ owner + "` team. Please update your permissions and try again.");
+         // need persistAndFlush otherwise we won't catch SQLGrammarException
+         test.persistAndFlush();
+      } catch (Exception e) {
+         throw ServiceException.serverError("Access change failed (missing permissions?)");
       }
    }
 
    @Override
    @RolesAllowed("tester")
```
johnaohara, line 497: Missed another `@RolesAllowed(Roles.TESTER)` here

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceNoRestTest.java
```diff
@@ -0,0 +1,639 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@QuarkusTest
+@QuarkusTestResource(PostgresResource.class)
+@TestProfile(HorreumTestProfile.class)
+@TestTransaction
+@TestSecurity(user = DEFAULT_USER, roles = {Roles.TESTER, Roles.VIEWER, FOO_TEAM, FOO_TESTER})
+class SchemaServiceNoRestTest extends BaseServiceNoRestTest {
```
johnaohara, line 43: For the tests that expect `ServiceExceptions`, I think we should check the response code to validate that the service is sending the correct http reponse as well as the error message

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceNoRestTest.java
```diff
@@ -0,0 +1,639 @@
...
+   }
+
+
+   @org.junit.jupiter.api.Test
+   void testCreateSchemaWithInvalidUri() {
+      String schemaUri = "dummy:schema";
+      Schema schema = createSampleSchema("Dummy schema", schemaUri, FOO_TEAM);
+
+      ServiceException thrown = assertThrows(ServiceException.class, () -> schemaService.add(schema));
+      assertEquals("Please use URI starting with one of these schemes: [urn, uri, http, https, ftp, file, jar]", thrown.getMessage());
```
johnaohara, line 97: In relation to the previous comment about the Exception message, something like 

```suggestion
      assertEquals(Response.Status.BAD_REQUEST.getStatusCode();, thrown.getResponse().getStatus());
      assertEquals("Please use URI starting with one of these schemes: [urn, uri, http, https, ftp, file, jar]", thrown.getMessage());
```

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceNoRestTest.java
```diff
@@ -0,0 +1,639 @@
...
+   // utility to create a schema in the db, tested with testCreateSchema
+   private Schema createSchema(String name, String uri, String token) {
+      Schema schema = new Schema();
+      schema.owner = FOO_TEAM;
+      schema.name = name;
+      schema.uri = uri;
+      schema.token = token;
+      schema.id = schemaService.add(schema);
+      return schema;
+   }
```
johnaohara, line 638: What is the difference between this and `BaseServiceNoRestTest.createSampleSchema()`?
Instead of duplicating code, we should use the method in the parent class

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TestServiceNoRestTest.java
```diff
@@ -0,0 +1,504 @@
...
+      assertNull(test4.folder);
+   }
+
+   @org.junit.jupiter.api.Test
+   void testCreateTestWithBlankName() {
+      Test t = createSampleTest("", null, null, null);
+
+      // empty test name
+      ServiceException thrown = assertThrows(ServiceException.class, () -> testService.add(t));
+      assertEquals("Test name can not be empty", thrown.getMessage());
```
johnaohara, line 98: Same comment I made in `SchemaServiceNoRestTest` about testing the Response http code against expected code

johnaohara (changes requested):

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
lampajr: Good to hear that, I think there are many other places (in other services) where we could remove them. I will try do that as part of other testing improvements :rocket:

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
lampajr: That's true, but actually that case was not working at all. 

It was setting `existing = label` and then Hibernate would have thrown an exception as the entity it was trying to persist was not managed/associated to the Hibernate session (I think because the id should not be set to `-1` but to `null`).

Given that it was not working before, IMO to be coherent wrt other services if the provided id is not null and there is no entity with that id we should simply fail. Otherwise if we have user wanting to update a label, set the wrong id, it would end up creating a new label which is not the intended behavior. wdyt?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
lampajr: Created --> https://github.com/Hyperfoil/Horreum/issues/1670

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
lampajr: Good catch, thx!

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceNoRestTest.java
lampajr: The only difference is line `636`, `createSampleSchema` simply creates the DTO object.. whereas `createSchema` creates a DTO and call the service to add it to the db.

I am refactoring this to make use of `createSampleSchema` inside `createSchema` body, something like:
```java
   // utility to create a schema in the db, tested with testCreateSchema
   private Schema createSchema(String name, String uri, String token) {
      Schema schema = createSampleSchema(name, uri, FOO_TEAM, token);
      schema.id = schemaService.add(schema);
      return schema;
   }
```

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceNoRestTest.java
lampajr: I agree, going to add those checks :+1:

lampajr:
Hi @johnaohara , thanks for the review :pray: 
I think I addressed all your comments (and applied changes accordingly) - see also this https://github.com/Hyperfoil/Horreum/pull/1656#discussion_r1595321834

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceNoRestTest.java
johnaohara: @lampajr have you added these checks?

lampajr: yes, see as an example: https://github.com/Hyperfoil/Horreum/pull/1656/files#diff-8f4fde724167cf6f29ef5a003807126facd504f1158a9cd6462c1be599242d81R116

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/SchemaServiceNoRestTest.java
lampajr: Added for all tests that were checking the `ServiceException`

johnaohara: great, I can see the changes now, might have been looking at an old version of the files :/

johnaohara (approved):

lampajr:
> @lampajr please can you squash and then I will merge, thanks

Done :heavy_check_mark:

PR #1689, opened 2024-05-13
Validate test/schema import format: Fixes #1685

### LINKED ISSUE #1685: Schema and Test import do not validate structure of JSON passed into API
**Describe the bug**

When a schema or test is imported, the mapping of json to the import configuration object(s) is not validated, allowing incorrectly structured JSON to be processed for import.

This was one part of the issue that lead to: https://github.com/Hyperfoil/Horreum/issues/1661

johnaohara:
## Fixes Issue
Fixes #1685

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
@@ -1,5 +1,10 @@
 package io.hyperfoil.tools.horreum.svc;
 
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationConfig;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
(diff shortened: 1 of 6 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/SchemaServiceImpl.java
```diff
@@ -849,7 +854,17 @@ public SchemaExport exportSchema(int id) {
    @Transactional
    @Override
    public void importSchema(ObjectNode node) {
-      SchemaExport importSchema = Util.OBJECT_MAPPER.convertValue(node, SchemaExport.class);
+      SchemaExport importSchema;
+
+      ObjectMapper customMapper = mapper.copy();
+      customMapper.enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
```
lampajr, line 860: Can we create this additional mapper only once and then simply inject this mapper? Otherwise we are going to create `mapper.copy();` every time we need to import something

johnaohara: I did think about that, but i was being lazy ;) thinking that importing schema/test is not on the hot path

lampajr: yeah, that's true.. I was thinking the same

lampajr (approved):

johnaohara:
@lampajr I just pushed another commit that produces a custom mapper, which is cached and reused as a CDI bean by default: https://github.com/Hyperfoil/Horreum/pull/1689/commits/7463f9409a65b6234d9f8cdc9474d99c42ae1215

lampajr (approved):
> @lampajr I just pushed another commit that produces a custom mapper, which is cached and reused as a CDI bean by default: [7463f94](https://github.com/Hyperfoil/Horreum/commit/7463f9409a65b6234d9f8cdc9474d99c42ae1215)

Thanks @johnaohara , lgtm!

PR #1703, opened 2024-05-15
Add Collector API datastore

### LINKED ISSUE #1109: Feature: Datastore integration with `Collector API`
An external Collector API tool (https://github.com/Karm/collector) allows performance metrics to be pushed and queried as JSON documents via a REST API

In order to integrate existing datasets into Horreum for analysis, a new `Collector API` datastore needs to be created that allows Horreum Users to query and existing instance.

The `Collector API` datastore should be defined as a new Datastore Type (https://github.com/Hyperfoil/Horreum/blob/master/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/Datastore.java), similar to how the Elasticsearch datastore is integrated (https://github.com/Hyperfoil/Horreum/blob/master/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/ElasticsearchDatastore.java)

Documentation on how the Elasticsearch document store is used can be found here: https://horreum.hyperfoil.io/docs/tutorials/query-elasticserach/

zakkak:
## Fixes Issue

Closes #1109 

## Proposed Changes

This patch issues a request to the collector API to get all available data. To limit the size of returned data we allow the use of a specific `tag` and `imgName` combination, while also limiting the date range.


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/CollectorApiDatastore.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+            metaData = payload;
+
+            final CollectorApiDatastoreConfig jsonDatastoreConfig = mapper.treeToValue(configuration.configuration, CollectorApiDatastoreConfig.class);
+
+            if ( jsonDatastoreConfig != null ){
+
+                HttpClient client = HttpClient.newHttpClient();
+                HttpRequest.Builder builder = HttpRequest.newBuilder()
+                        .uri(java.net.URI.create(jsonDatastoreConfig.url + "/api/v1/image-stats/tag/" + payload.get("tag").asText()));
```
johnaohara, line 51: should the url `path` be a separate config param, or part of the `jsonDatastoreConfig.url`, if the api changes (e.g. a version bump) then this hard-coded path will likely be no longer valid

johnaohara (changes requested):
One thing we need to think about is duplicated runs. As it currently works, the call to the collector api simply returns all the results for a given `tag`, each invocation will return all the data and will be uploaded into Horreum

So, we either need to be able to query the api to return a subset of the data, or we need to handle de-duplication in Horreum

johnaohara:
Ah, I now see this issue wrt filtering the api result: https://github.com/Karm/collector/pull/34

zakkak:
https://github.com/Karm/collector/pull/35 is now merged enabling filtering by tag, benchmark, and date range (not pushed to production yet though). So I will adapt the PR to use the new API.

johnaohara:
@zakkak excellent news.

zakkak:
@johnaohara this should now be ready for review.

johnaohara (approved):
LGTM!

johnaohara:
@zakkak Many thanks for your contribution!

PR #1708, opened 2024-05-16
Create initial account when there are no accounts in the system

barreiro:
This PR adds a mechanism that generates an account that can be used to bootstrap Horreum. That account should be removed once there are other accounts in the system.

Its another step towards removing dependency on keycloak and also custom dev-services. 

fixes https://github.com/Hyperfoil/Horreum/issues/1719

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityBootstrap.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityBootstrap.java
@@ -1,47 +1,60 @@
 package io.hyperfoil.tools.horreum.server;
 
+import io.hyperfoil.tools.horreum.api.internal.services.UserService;
 import io.hyperfoil.tools.horreum.entity.user.Team;
 import io.hyperfoil.tools.horreum.entity.user.TeamMembership;
 import io.hyperfoil.tools.horreum.entity.user.TeamRole;
 import io.hyperfoil.tools.horreum.entity.user.UserInfo;
 import io.hyperfoil.tools.horreum.entity.user.UserRole;
 import io.hyperfoil.tools.horreum.svc.Roles;
+import io.hyperfoil.tools.horreum.svc.user.UserBackEnd;
+import io.quarkus.logging.Log;
 import io.quarkus.runtime.StartupEvent;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
-import org.jboss.logging.Logger;
 import org.keycloak.admin.client.Keycloak;
 import org.keycloak.representations.idm.RoleRepresentation;
 import org.keycloak.representations.idm.UserRepresentation;
 
+import java.security.SecureRandom;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
-@ApplicationScoped
-public class SecurityMigration {
+import static io.quarkus.runtime.configuration.ProfileManager.getLaunchMode;
 
-    private static final Logger LOGGER = Logger.getLogger("SecurityMigration");
+@ApplicationScoped public class SecurityBootstrap {
 
     @ConfigProperty(name = "quarkus.keycloak.admin-client.server-url") Optional<String> keycloakURL;
-    @ConfigProperty(name = "quarkus.keycloak.admin-client.realm", defaultValue = "horreum") String realm;
+    @ConfigProperty(name = "horreum.keycloak.realm", defaultValue = "horreum") String realm;
 
     @ConfigProperty(name = "horreum.roles.provider", defaultValue = "keycloak") String provider;
 
+    @ConfigProperty(name = "horreum.bootstrap.password") Optional<String> providedBootstrapPassword;
+
     private static final String MIGRATION_PROVIDER = "database";
+    private static final String BOOTSTRAP_ACCOUNT = "horreum.bootstrap";
+
+    private static final char[] RANDOM_PASSWRORD_CHARS = ("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789").toCharArray();
+    private static final int RANDOM_PASSWORD_DEFAULT_LENGTH = 16;
 
     @Inject RoleManager roleManager;
 
+    @Inject Instance<UserBackEnd> backend;
+
     void onStart(@Observes StartupEvent event, Keycloak keycloak) {
         if (keycloakURL.isPresent() && performRolesMigration()) {
-            LOGGER.info("Perform roles migration from keycloak...");
+            Log.info("Perform roles migration from keycloak...");
             for (UserRepresentation kcUser : keycloak.realm(realm).users().list(0, Integer.MAX_VALUE)) {
                 performUserMigration(kcUser, keycloak.realm(realm).users().get(kcUser.getId()).roles().getAll().getRealmMappings());
             }
-            LOGGER.info("Migration from keycloak complete");
+            Log.info("Migration from keycloak complete");
         }
+        checkBootstrapAccount();
     }
 
     private boolean performRolesMigration() {
@@ -57,10 +70,10 @@ public class SecurityMigration {
 
     @Transactional
     void performUserMigration(UserRepresentation kcUser, List<RoleRepresentation> kcRoles) {
-        LOGGER.infov("Migration of user {0} {1} with username {2}", kcUser.getFirstName(), kcUser.getLastName(), kcUser.getUsername());
+        Log.infov("Migration of user {0} {1} with username {2}", kcUser.getFirstName(), kcUser.getLastName(), kcUser.getUsername());
         String previousRoles = roleManager.setRoles(kcUser.getUsername());
         try {
-            
+
             Optional<UserInfo> storedUserInfo = UserInfo.findByIdOptional(kcUser.getUsername());
             UserInfo userInfo = storedUserInfo.orElseGet(() -> new UserInfo(kcUser.getUsername()));
             userInfo.email = kcUser.getEmail();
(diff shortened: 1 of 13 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityInit.java
```diff
@@ -1,46 +1,53 @@
...
 
+import java.security.SecureRandom;
 import java.util.List;
 import java.util.Optional;
 
-@ApplicationScoped
-public class SecurityMigration {
-
-    private static final Logger LOGGER = Logger.getLogger("SecurityMigration");
+@ApplicationScoped public class SecurityInit {
```
barreiro, line 27: Renamed `SecurityMigration` to `SecuritiInit` as the initial account logic needs to occur *after* any migration (if it occurs).

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityInit.java
```diff
@@ -95,4 +102,29 @@ private void addTeamMembership(UserInfo userInfo, String teamName, TeamRole role
...
+     * Create an admin account if there are no accounts in the system.
+     * The account should be removed once other accounts are created.
+     */
+    @WithRoles(extras = Roles.HORREUM_SYSTEM)
+    void checkInitialAccount(UserBackEnd backend) {
+        long userCount = UserInfo.count();
+        if (userCount == 0) {
+            UserService.NewUser user = new UserService.NewUser();
+            user.user = new UserService.UserData("", INITIAL_ACCOUNT, "Temporary", "Admin", "temp@horreum.com");
+            user.password = new SecureRandom().ints(RANDOM_PASSWORD_LENGTH, '0', 'z' + 1).collect(StringBuilder::new, StringBuilder::appendCodePoint, StringBuilder::append).toString();
```
barreiro, line 118: generates a password from random characters from ASCII `0` to `z` (includes digits, upper and lowercase letters and punctuation --- all characters in the range are printable)

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityInit.java
```diff
@@ -95,4 +102,29 @@ private void addTeamMembership(UserInfo userInfo, String teamName, TeamRole role
...
+
+            backend.createUser(user);
+            backend.updateAdministrators(List.of(INITIAL_ACCOUNT));
+            Log.infov("Created temporary admin account: {0} with password {1}", INITIAL_ACCOUNT, user.password);
+
+        } else if (userCount > 1 && !backend.info(List.of(INITIAL_ACCOUNT)).isEmpty()) {
+            Log.warnv("Please remove temporary admin account: {0}", INITIAL_ACCOUNT);
+            // TODO: remove account or add a way to remove accounts
```
barreiro, line 126: at the moment there is no way to remove an account. would need to implement that feature.

stalep:
@barreiro again, which issue is this solving?

barreiro:
fixes #1719

johnaohara:
@barreiro please can you take a look at the test failures?

stalep:
@barreiro does these changes require any new/changes documentation?

barreiro:
@stalep yes, but I'm planning to do it in a separate PR that addresses all the changes in authentication.

johnaohara:
@barreiro if you are planning on adding documentation to cover authentication, please can you add an issue so that we do not forget and can prioritise bringing the docs in line with functionality.  Ideally everything (docs, test, changes etc) should be in on PR

barreiro:
@johnaohara created #1775 to follow up with documentation.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityBootstrap.java
```diff
@@ -95,4 +111,33 @@ private void addTeamMembership(UserInfo userInfo, String teamName, TeamRole role
...
+    public void checkBootstrapAccount() {
+        // checks the list of administrators. a user cannot remove himself nor create the bootstrap account (restricted namespace)
+        List<String> administrators = backend.get().administrators().stream().map(userData -> userData.username).toList();
+        if (administrators.isEmpty()) {
+            UserService.NewUser user = new UserService.NewUser();
+            user.user = new UserService.UserData("", BOOTSTRAP_ACCOUNT, "Bootstrap", "Acount", "horreum@example.com");
+            user.password = providedBootstrapPassword.orElseGet(() -> getLaunchMode().isDevOrTest() ? "secret" : generateRandomPassword(RANDOM_PASSWORD_DEFAULT_LENGTH));
+
+            backend.get().createUser(user);
+            Log.infov("Created temporary account {0} with password {1}", BOOTSTRAP_ACCOUNT, user.password);
```
johnaohara, line 130: I think this needs to be more prominent in the log
```suggestion
            Log.infov("\n\n*********************************\nCreated temporary account `{0}` with password `{1}`\n*********************************", BOOTSTRAP_ACCOUNT, user.password);
```

creates the following output:

```
2024-06-13 16:05:41,982 INFO  [io.sma.rea.mes.amqp] (vert.x-eventloop-thread-1) SRMSG16203: AMQP Receiver listening address dataset-event
2024-06-13 16:05:42,306 INFO  [io.hyp.too.hor.ser.SecurityBootstrap] (Quarkus Main Thread) 

*********************************
Created temporary account `horreum.bootstrap` with password `secret`
*********************************

2024-06-13 16:05:42,389 INFO  [io.hyp.too.hor.svc.use.KeycloakUserBackend] (Quarkus Main Thread) Added administrator role to user horreum.bootstrap

```

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityBootstrap.java
```diff
@@ -1,46 +1,62 @@
...
-            LOGGER.info("Perform roles migration from keycloak...");
+            Log.info("Perform roles migration from keycloak...");
             for (UserRepresentation kcUser : keycloak.realm(realm).users().list(0, Integer.MAX_VALUE)) {
                 performUserMigration(kcUser, keycloak.realm(realm).users().get(kcUser.getId()).roles().getAll().getRealmMappings());
             }
-            LOGGER.info("Migration from keycloak complete");
+            Log.info("Migration from keycloak complete");
+        }
+
+        if (getLaunchMode() != LaunchMode.TEST) { // test client woul need the 'realm-admin' role {@link io.hyperfoil.tools.horreum.svc.KeycloakUserServiceTests}
```
johnaohara, line 58: Do we need the check for Test mode? the default username/password is going to be the same for dev and test, so can't we just rely on that fact for tests?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityBootstrap.java
```diff
@@ -1,46 +1,62 @@
...
+    @ConfigProperty(name = "horreum.keycloak.realm", defaultValue = "horreum") String realm;
 
     @ConfigProperty(name = "horreum.roles.provider", defaultValue = "keycloak") String provider;
 
+    @ConfigProperty(name = "horreum.bootstrap.password") Optional<String> providedBootstrapPassword;
+
     private static final String MIGRATION_PROVIDER = "database";
+    private static final String BOOTSTRAP_ACCOUNT = "horreum.bootstrap";
+
+    private static final char[] RANDOM_PASSWRORD_CHARS = ("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!@#$%^*()-_=+[{]}|;:,<.>").toCharArray();
```
johnaohara, line 42: There are some special chars that cause passwords to fail when authenticating via curl. There was an instance today where a password containing special chars works via the UI, but does not work with CURL. 
I have not had chance to investigate which chars caused auth via curl to fail, but it this will create a random password that is stored in keycloak, i suggest we restrict the special chars from password

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
```diff
@@ -149,7 +149,7 @@ private static UserRepresentation convertUserRepresentation(UserService.NewUser
 
         CredentialRepresentation credentials = new CredentialRepresentation();
         credentials.setType(CredentialRepresentation.PASSWORD);
-        credentials.setTemporary(true);
+//        credentials.setTemporary(true);
```
johnaohara, line 152: if this is no longer required, please remove

johnaohara (changes requested):
By default, In dev mode, there is a `dev-team` created by default.  This PR provides the initial admin user in dev mode and prod mode, but removes the default team.

We will need;

-  docs to walk users through configuring Horreum before it can be used, e.g. creating a team, users and assigning them to the new team 

ideally, we would also provide a wizard to walk new admins through through the process of creating new teams at first startup

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityBootstrap.java
barreiro: I have already reduced the set of symbols, but let's stick with alphanumeric

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
barreiro: removed. it was required, but I was now able to work around it,

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityBootstrap.java
barreiro: makes sense. I changed to something similar (3 lines in the output)

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/SecurityBootstrap.java
barreiro: I was able to make this work, by adding `HorreumKeycloakTestResourceLyfecycleManager` to wrap `KeycloakTestResourceLyfecycleManager` to setup the OIDC client the right way. 

Most tests run with the `HorreumTestProfile` that uses WireMock to provide authentication, and that obviously does not allow to create/remove users. Luckly (or not :wink:) Horreum can work with with an external OIDC provider :smile:  so I changed the testsuite to run that way (`horreum.roles.provider=database` and `horreums.roles.database.override=false`)  to create the bootstrap account in the DB.

The upside of this work is the consistency between all profiles.

barreiro:
@johnaohara I have addressed your comments but the one about the default team. I believe creating such a team is part of initial setup, but is not a requirement.

I'll take that into account in #1775

johnaohara:
> @johnaohara I have addressed your comments but the one about the default team. I believe creating such a team is part of initial setup, but is not a requirement.
> 
> I'll take that into account in #1775

Dev mode is about developer productivity. If developers have to go through manual steps to recreate a team and assign a user every time they start horreum in dev mode, that is a regression wrt productivity.

right now, i can start dev mode and immediately create a job and start pushing data into Horreum.  I would even argue that we should have a default schema and test so that I do not have to complete those steps.

We can certainly have a conversation about removing the default team, but I think we need to have a conversation about what we think the behaviour *should* be and make changes that improves user/dev experience rather than tagging this in a PR that creates a default initial user.

Options *could* be; 

 1 - a wizard that pops up when Horreum does not contain any teams and asks the user if they want to load dummy data or walks them through creating a new team
 2 - a dev mode option to persist the database between restarts, allowing developers to only create the team once on the first startup, but re-use the same database files, containing the same configuration

But lets open an issue and discuss, rather than removing it here and debating in a PR that is linked to an issue that does not refer to removing the default team in dev mode

barreiro:
ok, fair enough. I don't have a very strong opinion about it. I do agree with your long term plans, specially number two. agree that we need an issue to track it.

johnaohara (approved):

PR #1710, opened 2024-05-20
add multiFilter to run and test labelValues

### LINKED ISSUE #1695: /labelValue?filter multi-select filtering for Grafana
filters were created for 2 use cases:
1. Does the provided filter object exist in the label value map? This allows user to pass in a "subset" of the label value map and filter for only those maps that match. This is akin to passing in a fingerprint object and only getting back label value maps that contain the fingerprint.
2. Does the provided jsonpath filter match the label value map? e.g. `$."Number of Samples" ? (@ > 50)`. This is a more powerful option but more complicated and is directly tied to the implementation of our jsonpath processing (postgres) so I do not think we should advertise this option to users atm.

Option 1 works with grafana filtering when looking for specific fingerprints. It does NOT work with grafana variable based fintering in 2 situations:
1. Selecting all values.
if label "cpu" has possible values 1,2,4,8 then grafana's default option to select all values using variables would be in included `"cpu": [1,2,4,8]` (assuming json encoding). Horreum would see this as a label with an array value and not an array of possible values
Grafana offers custom encoding for "All" values (e.g. `*`) but at the moment Horreum would see that a string value for the "cpu" label.
Horreum's current method to select All values would be to exclude the label name from the fingerprint. Unfortunately Grafana does not appear to offer that option when configuring filtering.
2. Selecting multiple values.
Similar to the default all values, Grafana will encode selecting some of the values as an array of options. e.g. `"cpu": [4,8]`. Horreum would see this as a label with an array value and not an array of possible values.

Options:

I believe Horreum must support Grafana multi-select if it will be considered a viable solution for user charting.

* We can support the "all values" problem relatively easily by checking for a custom encoded value (e.g. `*`) and removing that from the filter object. This will prevent all label's from using that value as a valid label value. I don't like the idea of doing this but it is admittedly easier and faster to implement. 
The problem is that this is a "Forever" decision or a breaking api change whenever / if ever we revert it. I personally think we should not introduce this option because it is only a partial solution (doesn't address multiple value selection) and it places value restrictions on all labels.

* We could similarly use graphite encoding ( `"cpu": {1,2,4,8}`) but inside a string ( `"cpu": "{1,2,4,8}"` ) and use the same custom filter from above to determine this is an array of options and not the desired content of a string. This along with some indicator paramater to "enable grafana parameter parsing" would be able to support both all and multi-select. One downside is that it removes the possible label value (probably a minor detail) and it mixes model representation. I am not a fan of including json and grafite because it is tightly coupled with grafana. The next UI charting tool (jupyter, tablua, etc) would have a different query syntax and we would end up with a growing list of integration support.

The next option is to introduce a domain specific filtering syntax that is compatible with Grafana. This would need to support how Grafana encodes single `"cpu": 4`, multiple `"cpu": [1,8]`, and all variable options `"cpu" : [1,2,4,8]` or `"cpu":"*"`. 
The challenge is to differentiate arrays that represent value options versus label values that are arrays. 

I looked at the domain languages used by mongo or elastic but they both make use of Grafana plugins to convert the variable input into their query syntax.

A naive solution to this could be to perform both queries and return the longest list of results. The theory being the incorrect interpretation would return 0 results (checking for a equality filter object is a multi-select vs checking for containment when the label value is an array)
This is not a good choice because it requires 2x the number of arrays in the filter to find the correct solution. That will hopefully usually be small but is still quite wasteful.

willr3:
Closes #1695

This is a simple implementation adding `multiFilter` to the `/labelValues` endpoints so that Horreum will check for each entry in a filter array instead of looking for a matching array.


horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/RunServiceTest.java
```diff
@@ -108,29 +110,87 @@ public void testTransformationWithoutSchema(TestInfo info) throws InterruptedExc
...
       assertNewDataset(dataSetQueue, runId);
    }
-   private int labelValuesSetup(Test t) throws JsonProcessingException {
+   private int loadRun(String json,String uri,String testName) throws JsonProcessingException {
+      ObjectMapper mapper = new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);;
+      JsonNode data = mapper.readTree(json);
+      String idString = uploadRun(data,testName,uri);
+      int id = Integer.parseInt(idString);
+      return id;
+   }
```
lampajr, line 119: Given that we are defining the same `loadRun` method in both `RunServiceTest` and `TestServiceTest` (and the implemenation looks the same), I would suggest to move the method in the `BaseServiceTest` to avoid duplication

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
```diff
@@ -274,16 +275,52 @@ public Object getData(int id, String token, String schemaUri) {
...
+               filterSql = "WHERE ";
+            }
+            for(int i=0; i<assumeMulti.size(); i++){
+               if(!filterSql.endsWith("WHERE ")){
+                  filterSql+=" AND ";
+               }
+               filterSql += " jsonb_path_query_first(combined.values,CAST( :key"+i+" as jsonpath)) = ANY(:value"+i+") ";
+            }
+         }
+
```
lampajr, line 323: I see this new logic is duplicated in `TestServiceImpl` (the only difference I spot seems that, here, we are not defining `FILTER_PREFIX`), can we extract it in a common place to avoid duplication? I think this will make further fixes easier to do if we need to do in one single place ratehr than two.

lampajr (changes requested):
Hi @willr3, 

Looks like we need to re-generate the openapi spec (it should be enough to run `mvn package` and commit the additional changes)

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/RunServiceTest.java
willr3: Yep, that one is probably good to move into the `BaseServiceTest` as an alternative form of `uploadRun`

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
johnaohara: +1 I think we should avoid duplication

willr3: I agree, the sql manipulation should be done in one place. The reason I didn't is because all of this code has to be replaced when we transition to composable labels. I originally only created the `multiFilter` for the `test` `labelValues` endpoint but was asked to also add it to `run` for "consistency". The expectation is that this code will be gone in a few weeks but that is probably overly optimistic. I'll refactor :)

lampajr (changes requested):
Added some other comments below, actually just minors.

You still need to re-generate the OpenAPI spec by running `mvn package` otherwise the CI keeps failing on the `check uncommitted changes`:
```bash
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -1043,6 +1043,13 @@ paths:
           multiple:
             description: excluding multiple labels
             value: "id,count"
+      - name: multiFilter
+        in: query
+        description: enable filtering for multiple values with an array of values
+        schema:
+          default: false
+          type: boolean
+        example: true
       responses:
         "[20](https://github.com/Hyperfoil/Horreum/actions/runs/9168643433/job/25213861325?pr=1710#step:9:21)0":
           description: label Values
@@ -[21](https://github.com/Hyperfoil/Horreum/actions/runs/9168643433/job/25213861325?pr=1710#step:9:22)55,6 +2162,13 @@ paths:
           multiple:
             description: excluding multiple labels
             value: "id,count"
+      - name: multiFilter
+        in: query
+        description: enable filtering for multiple values with an array of values
+        schema:
+          default: false
+          type: boolean
+        example: true
       responses:
         "200":
           description: OK
```

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
```diff
@@ -274,37 +275,27 @@ public Object getData(int id, String token, String schemaUri) {
...
+      TestServiceImpl.FilterDef filterDef = TestServiceImpl.getFilterDef(filter,null,null,multiFilter,(str)->
+              labelValues(runId,str,sort,direction,limit,page,include,exclude,false),em);
+
+      String filterSql = filterDef.sql();
+      if(filterDef.filterObject()!=null){
+         filterObject = filterDef.filterObject();
       }
+
       if(filterSql.isBlank() && filter != null && !filter.isBlank()){
          //TODO there was an error with the filter, do we return that info to the user?
```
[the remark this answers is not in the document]
willr3: I agree the API could use defined error handling but I think that is out of scope for this change.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
```diff
@@ -349,12 +341,21 @@ SELECT DISTINCT COALESCE(jsonb_object_agg(label.name, lv.value) FILTER (WHERE la
       if(!filterSql.isEmpty()) {
```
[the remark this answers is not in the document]
willr3: I think error handling is good to discuss in another issue. I think this PR should focus on the added `multiFilter` functionality.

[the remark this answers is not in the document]
lampajr: yeah makes sense, I thought it was part of the changes :facepalm:

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -602,28 +650,59 @@ public List<ExportedLabelValues> labelValues(int testId, String filter, String b
...
+      String orderSql = "";
+
+      Instant beforeInstant = Util.toInstant(before);
+      Instant afterInstant = Util.toInstant(after);
+
+      FilterDef filterDef = getFilterDef(filter,beforeInstant,afterInstant,multiFilter,(str)->
+         labelValues(testId, str, before, after, filtering, metrics, sort, direction, limit, page, include, exclude, false)
+      ,em);
+
+      String filterSql = filterDef.sql();
```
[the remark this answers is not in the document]
lampajr: Resolving as the check is performed on line 746

lampajr (approved):
lgtm

johnaohara (approved):

Issue #1674, opened 2024-05-10
PROTECTED tests are saved as PRIVATE ones


lampajr:
**Describe the bug**

Trying to create a `PROTECTED` test ends up in creating a `PRIVATE` one.

After some investigation looks like a UI issue which is sending the wrong access field while `PROTECTED` is selected from the user.

**To Reproduce**

1. Create a new `Test` with `PROTECTED` access
2. Check the newly created test in the UI, it will show `PRIVATE` access
![image](https://github.com/Hyperfoil/Horreum/assets/26715795/94d83afd-b14e-4223-9836-b1b6274f3bb3)
3. Checking the API request looks like the UI is sending `PRIVATE` instead of `PROTECTED`
![image](https://github.com/Hyperfoil/Horreum/assets/26715795/78c3b8e1-fc2d-473f-9a47-67df8a364260)

**Version**

What is the version of Horreum ?
```bash
$ git rev-parse HEAD
2921d7fcde0e4830fecbbf2aef3fdb229f315a81
```

lampajr:
Looks like the issue is in how the UI has been structured, for new Tests only two tabs are visibile: `Setting` and `Access`:
![image](https://github.com/Hyperfoil/Horreum/assets/26715795/f6973355-d3c4-4b26-8397-c81bceba7ab2)

Then, 
1. `Save` is working only in the `Settings` tab
2. The two tabs are not **sharing** the state, therefore any change made on `Access` tab is not visible in the `Settings` one
3. Saving the test will send all information set in `Settings` tab and completely ignoring those in `Access` tab like `owner`, `token` and `access rights`

As a result, you cannot create a `Test` with:
1. `PUBLIC` or `PROTECTED` access right as the default is `PRIVATE`
2. A custom access token
3. An owner different from the default one

PR #1720, opened 2024-05-21
Refactor test settings and access rights

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/1674

## Changes proposed

In order to properly fix the issue, I had to slightly refactor the UI for Test settings and access rights.

This is the list of changes I did:
- [x] Moved the `Access rights` tab into the `Settings` one, this way you can set the proper permissions during the `Test` creation 
- [x] Kept `Test Tokens` creation/deletion in a separate tab (which will replace the `Access` one, now called `Tokens`)
- [x] Refactored the `Tokens` management to be consistent with other UI screens
- [x] Moreover, once tokens are update the changes are automatically reflected in the UI without any reload page
- [x] The new `Tokens` tab is NOT visible during `Test` creation as the backend API requires an existing Test to properly add/delete tokens. 

With this new UI, users are able to create a new `Test` with custom permissions, either `Public` or `Protected` (which is something not working before)

Here a brief summary of UI changes in the form of before/after:

**Settings**

Before:
image

After:
image

**Tokens**

Before:
image
image
image

After:
image
image
image


lampajr:
> @lampajr when I am testing this with a backup of production, have you seen these when you have tested this change?

Mmmh I do not remember any error like that, especially because I did not touch the backend :thinking: but let me check if I can reproduce the same.

lampajr:
I can see the request is performed, but it looks working on my side:
image

Are you experiencing in some specific situation? On my side I tried creating new test and updating those settings and tokens,

johnaohara:
> > @lampajr when I am testing this with a backup of production, have you seen these when you have tested this change?
> 
> Mmmh I do not remember any error like that, especially because I did not touch the backend 🤔 but let me check if I can reproduce the same.

I see the same issue in master, please ignore the exception

horreum-web/src/domain/tests/TestSettings.tsx
```diff
@@ -92,25 +101,25 @@ export default function TestSettings({ test, onTestIdChange, onModified, funcsRe
                 notificationsEnabled,
                 fingerprintLabels: [],
                 fingerprintFilter: undefined,
```
johnaohara, line 103: With the modified UI,  changes to managing access are now sent sent via POST to `/api/test`, with the full test data.

These defaults are dropping `fingerprintLabels` and `fingerprintFilter` from the existing test, causing fingerprints to be re-calculated for the test

lampajr: Yeah because with the change, managing access would be like managing any other Test setting.

Do you see a problem in recalculating fingerprints even if we are just changing access rights? I mean I am not expecting many access rights changes on a test so I don't think that could be an issue, wdyt?

johnaohara: Yes, if you select a test that has change detection variables set with a fingerprint label and change Access, the fingerprint labels are dropped and all the fingerprints are recalculated

lampajr: but isn't that the actual behavior if you change the Test setting?

lampajr: because if that was an issue also before (changing the settings will trigger the fingerprint recalculation, I still don't get if it was intended or not) I can easily fix that with something like:
```js
                fingerprintLabels: test?.fingerprintLabels || [],
                fingerprintFilter: test?.fingerprintFilter,
```

johnaohara: Looking back, yes it is an existing bug, but I think we should fix it in this PR

lampajr: > but I think we should fix it in this PR

I agree, fixed exacty as I was suggesting above https://github.com/Hyperfoil/Horreum/pull/1720/files#diff-0596d0bd925fab40460a7fef8c88687a33fe10df0e2bf55ab77f92fab27eb9a5R102-R103

PR #1725, opened 2024-05-22
Only initialize graal JS context when required. Fixes #1723

### LINKED ISSUE #1723: A Graal JS context is initialized for every datapoint calculation, regardless if it is required
**Describe the bug**

For every datapoint calculation, a new graal js context is created, whether it is required or not.  This is a very expensive operation;

image

For this upload, where there were no combination functions defined, setting up the context took 34.57% of total cpu cycles

johnaohara:
## Fixes Issue

Fixes: #1723 

In the flamegraphs below, calls to `Util.setupContext()` are highlighted

Before;

image

After ;

image


lampajr (approved):
lgtm

Issue #1412, opened 2024-03-05
Provide Tests Filtering Mechanism in UI


lampajr:
**Feature idea.**

I think that a possible improvement from UI perspective, especially when you have many `Test` records, would be to add a functionality to query/filter/search `Tests` by `name`.

**Describe the solution you'd like**

We could add an input text in the main table where user can choose the `field` to filter (initially it could be just the `name`) and its value:
![image](https://github.com/Hyperfoil/Horreum/assets/26715795/c836716b-1213-4dc9-8043-9a2bd347c08d)

Or alternatively directly adding the filter feature at column level like:
![image](https://github.com/Hyperfoil/Horreum/assets/26715795/788e8642-b059-4196-8c35-4299371458a7)

**Additional information**

The same could be implemented for `Schemas`

lampajr:
When I created this issue I missed there was a similar one already: https://github.com/Hyperfoil/Horreum/issues/1198.

Let's keep this one as well since it is specific for `Test` entities filtering.

lampajr:
As part of this change, I am slightly refactoring the custom table component with the following changes:
* Switch to Patternfly Table component
* Fix some ordering issues
* Reduce backend calls

Issue #1198, opened 2024-01-29
Add a quick search/filter field to Schema and Tests page


jhutar:
**Feature idea.**

It would be great if there is a quick search/filter field on Schema (and Tests) page. Imagine I'm looking for `urn:rhtap-load-test:max-concurrency:1.0` and I need to paginate through 5 pages and use Ctrl+F or so on each of them.

**Describe the solution you'd like**

It would be great if there is a way how to quickly find schema you are looking for if you know part of it's name.

johnaohara:
On Hold due to https://github.com/Hyperfoil/Horreum/discussions/1603

lampajr:
Given that https://github.com/Hyperfoil/Horreum/discussions/1603 was paused, I think this improvement would be really helpful because searching for schemas is becoming really annoying as the number of schemas increase

PR #1731, opened 2024-05-27
Implement test filtering by name

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/1412

## Changes proposed

The main purpose of this change was to add a name filtering on AllTests page.
As part of this new feature, I had chance to optimize, clean and refactor all tables trying to remove code duplication as much as possible.

Client-side:
- [x] Rename `Table` to `CustomTable` to differentiate with the Patternfly one
- [x] Use patternfly 5 tables, rather than raw tables
- [x] Reduce number of internal calls to the backend when changing sorting (it was calling the same endpoint two times, thus refreshing the UI two times in a row)
- [x] Performing server-side ordering on AllTests page (in prod it is performing client-side ordering, which is actually a page-isolated ordering)
- [x] Fix Runs ordering by Datasets size (there was wrong accessor)
- [x] Removed patternfly react-deprecated Dropdowns for all table columns, e.g., `watching` and `actions`
- [x] Moved pagination component inside the `CustomTable` itself to avoid duplication

Server-side:
- [x] Added new query parameter named `name` on TestSummary endpoint
- [x] Added test on that method


lampajr:
Below you can find some of the changes, but I'd encourage you to give it a try locally as the changes impacted all places where a Table was/is used:

Filter by name:
image

Improved a little bit the dataset comparison by wrapping it into a PageSection:
image

lampajr:
The only doubt I have is whether we want to add the filtering on the Schemas as well (this would be appreciated, see https://github.com/Hyperfoil/Horreum/issues/1198). I know that all changes regarding the schemas have been put on hold, but I am not expecting huge effort to extend this mechanism to the Schema entities (especially considering that we don't know when we can start the schema refactoring). wdyt @johnaohara ?

horreum-web/src/components/CustomTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+    count?: number,
+    perPage?: number,
+    page?: number,
+    onSetPage?: OnSetPage
+    onPerPageSelect?: OnPerPageSelect,
+    variant?: 'top' | 'bottom' | PaginationVariant,
+    isCompact?: boolean
+}
+
+type CustomTableProps<D extends object> = {
```
johnaohara, line 38: Here we define a custom table that is using generics, but the only place we specify a type is in `horreum-web/src/domain/schemas/SchemaList.tsx`

If we are going to use generics to enforce type checking at compile time, I think we should extend the usage to cover all instances of `<CustomTable/>`

johnaohara (changes requested):
@lampajr the only major issue is the Changes modal popup

johnaohara:
> The only doubt I have is whether we want to add the filtering on the Schemas as well (this would be appreciated, see #1198). I know that all changes regarding the schemas have been put on hold, but I am not expecting huge effort to extend this mechanism to the Schema entities (especially considering that we don't know when we can start the schema refactoring). wdyt @johnaohara ?

Lets implement Schema name filtering in a separate PR

johnaohara:
@lampajr please can you squash commits?

lampajr:
> @lampajr please can you squash commits?

Squashed and applied all your suggestions!

horreum-web/src/components/CustomTable.tsx
lampajr: Yeah make sense, cleaner this way! 
Added on all usages :heavy_check_mark:

horreum-web/src/domain/alerting/ChangeTable.tsx
```diff
@@ -72,13 +84,13 @@ const ChangeMenu = ({ change, onDelete, onUpdate }: ChangeMenuProps) => {
...
-                onClose={() => setModalChange(undefined)}
-                onUpdate={onUpdate}
-            />
+                <ChangeModal
+                    change={modalChange}
+                    isOpen={!!modalChange}
+                    onClose={() => setModalChange(undefined)}
+                    onUpdate={onUpdate}
+                />
+            </Dropdown>
```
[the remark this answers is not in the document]
lampajr: I think it was my fault, I wrongly put the modal inside the dropdwon. Fixed now.

johnaohara (approved):
LGTM

PR #1742, opened 2024-05-29
Use RH_PERF_BOT_TOKEN for backporting

lampajr:
## Changes proposed

Use machine account @rh-appservices-perf token for backporting purposes.


lampajr:
Unofortunately I think the only way to check this is actually working is to see it in action the first time we need it.

johnaohara (approved):

PR #1774, opened 2024-06-13
Remove accounts

### LINKED ISSUE #1770: Allow removal of machine and user accounts accounts
**Describe the bug**

In pr https://github.com/Hyperfoil/Horreum/pull/1711 the functionality was added to create machine accounts.

Migrating away from keycloak to manage users now mean there is not the ability for system admins to remove users.

As an administrator, I should be able to remove existing user accounts and machine accounts from the system to revoke access, whether I am using a keycloak backend or a database backend

barreiro:
Fixes #1770


horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/internal/services/UserService.java
```diff
@@ -47,6 +47,11 @@ public interface UserService {
    @Blocking
    void createUser(@RequestBody(required = true) NewUser user);
 
+   @DELETE
+   @Path("removeUser")
+   @Blocking
+   void removeUser(@RequestBody(required = true) String username);
```
johnaohara, line 53: In a restful API, the delete method should "delete the resources (identified by the Request-URI)." So a call should be 
DELETE http://exmaple.com/api/user/anon

```suggestion
   @DELETE
   @Path("{username}")
   @Blocking
   void removeUser(@PathParam(name = "username", required = true) String username);
```

johnaohara (changes requested):

johnaohara (approved):

Issue #1719, opened 2024-05-21
Initial account


barreiro:
**Feature idea.**

There needs to exist one account available to bootstrap Horreum.

**Describe the solution you'd like**

I propose that on startup, when there are no accounts registered, create one on the reserved namespace `horreum.` with `admin` role. The account will have a random generated password (for dev / test can be a default  instead) that will be available in the startup logs It can be removed later on once another admin account is created.

**Additional information**

At the moment there is a pre-loaded account into the keycloak dev service. That 1) is not available in deployments without oidc and 2) does not answer the problem of bootstrap in production profile.

PR #1794, opened 2024-06-15
[0.13.x] Create initial account when there are no accounts in the system

rh-appservices-perf:
**Backport:** https://github.com/Hyperfoil/Horreum/pull/1708

This PR adds a mechanism that generates an account that can be used to bootstrap Horreum. That account should be removed once there are other accounts in the system.

Its another step towards removing dependency on keycloak and also custom dev-services. 

fixes https://github.com/Hyperfoil/Horreum/issues/1719

### DIFF
--- a/horreum-backend/pom.xml
+++ b/horreum-backend/pom.xml
@@ -152,10 +152,6 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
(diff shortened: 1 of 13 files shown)


johnaohara:
Backport has been opened against wrong stable branch

Issue #1346, opened 2024-02-19
Dataset filtering should use the patternfly design recommendation


willr3:
**Feature idea.**

Dataset filtering in the UI uses a drop down per filter option (labelName?) which can result in a long scrollable header of options

![Screenshot from 2024-02-19 12-16-06](https://github.com/Hyperfoil/Horreum/assets/1083859/d2c68e80-a703-4727-9c00-1b3d63473ef3)

The Patternfly recommended approach is the `attribute-search` example in the filters component

https://www.patternfly.org/patterns/filters#attribute-search

PR #1797, opened 2024-06-16
Change dataset filters to Patternfly Attribute Filter pattern: Fixes: #1346

johnaohara:
This PR changes the filter dropdown to use the Patternfly `attribute-search` pattern

## Fixes Issue

Fixes:  #1346
Screenshot from 2024-06-16 13-52-43

Screenshot from 2024-06-16 13-52-46

### DIFF
--- a/horreum-web/src/components/LabelFilter/LabelFilterOption.tsx
+++ b/horreum-web/src/components/LabelFilter/LabelFilterOption.tsx
@@ -1,0 +1,111 @@
+import React from "react";
+import {Menu, MenuContent, MenuItem, MenuList, MenuToggle, Popper} from "@patternfly/react-core";
+
+interface FilterOptionProps {
+    name: string;
+    options: string[];
+    filter: any;
+    setFilter(filter: any): any;
+    selection: string;
+    setSelection(selection: string): any;
+}
+
+
+export default function LabelFilterOption(props: FilterOptionProps) {
+    const [isMenuOpen, setIsMenuOpen] = React.useState<boolean>(false);
+    const toggleRef = React.useRef<HTMLButtonElement>(null);
+    const menuRef = React.useRef<HTMLDivElement>(null);
+    const containerRef = React.useRef<HTMLDivElement>(null);
+
+    const handleMenuKeys = (event: KeyboardEvent) => {
+        if (isMenuOpen && menuRef.current?.contains(event.target as Node)) {
+            if (event.key === 'Escape' || event.key === 'Tab') {
+                setIsMenuOpen(!isMenuOpen);
+                toggleRef.current?.focus();
+            }
+        }
+    };
+
+    const handleClickOutside = (event: MouseEvent) => {
+        if (isMenuOpen && !menuRef.current?.contains(event.target as Node)) {
+            setIsMenuOpen(false);
+        }
+    };
+
+    React.useEffect(() => {
+        window.addEventListener('keydown', handleMenuKeys);
+        window.addEventListener('click', handleClickOutside);
+        return () => {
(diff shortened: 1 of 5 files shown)


horreum-web/src/components/CustomTable.tsx
```diff
@@ -138,7 +138,7 @@ function CustomTable<D extends object>({
 
     return (
         <Card>
-            {(title || pagination?.top) && (
+            {(title && pagination?.top) && (
```
lampajr, line 141: I don't think this should be an `&&` because there might be situation where you want the pagination without the title
```suggestion
            {(title || pagination?.top) && (
```

horreum-web/src/domain/runs/TestDatasets.tsx
```diff
@@ -208,10 +211,33 @@ export default function TestDatasets() {
...
-    return (
-        <>
-            <Toolbar className="pf-v5-u-justify-content-space-between" style={{ width: "100%" }}>
-                <ToolbarGroup>
+    const arrayOfClearCallbacks : any[] = [];
+    const clearCallback = (callback: () => void) => {
+        arrayOfClearCallbacks.push(callback);
+    }
+
+    // TODO: clear all filter cards
```
lampajr, line 219: Is this still a valid `TODO`?

horreum-web/src/components/LabelFilter/LabelFilterOption.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    const menu = (
+        <Menu ref={menuRef} id="attribute-search-{props.name}-menu" onSelect={onSelect} selected={props.selection}>
+            <MenuContent>
+                <MenuList>
+                    {
+                        props.options.map((option) => (
+                            <MenuItem isSelected={props.selection === option} itemId={option} key={option}>
+                                {option}
+                            </MenuItem>
```
lampajr, line 90: We need to filter out `undefined` options otherwise we would have null options (that are selected by default):
image
```suggestion
                        props.options.filter(option => option).map((option) => (
                            <MenuItem isSelected={props.selection === option} itemId={option} key={option}>
                                {option}
                            </MenuItem>
```

lampajr (changes requested):
Hey @johnaohara, this looks an awesome improvement in the UI :heart: 

I left a couple of comments.

horreum-web/src/domain/runs/TestDatasets.tsx
johnaohara: No, I have implemented this, I will remove the comment

horreum-web/src/components/LabelFilter/LabelFilterOption.tsx
johnaohara: The filter does remove out `undefined`, but the api (maybe it is the json marshalling) is returning a mixture of `null` and undefined, and it is the `null` values that are causing the issue. 

Other LabelValues only return `undefined` for an unknown value.

My intention was to merge this PR as it is, and then fix the api problem that is returning the wrong data type for an unknown label value

horreum-web/src/components/CustomTable.tsx
johnaohara: yes, i need to revert this and fix it correctly, there is an issue where the CardHeader is being displayed without a title or pagination

horreum-web/src/components/LabelFilter/LabelFilterOption.tsx
lampajr: Got it :+1:

lampajr (approved):
LGTM

PR #1808, opened 2024-06-20
 Reduce the amount of calls to `RoleManager.setRoles`

### LINKED ISSUE #1726: `RoleManager.setRoles()` called multiple times during a transaction
**Describe the bug**

`RoleManager.setRoles()` is a very expensive operation, which requires a full round trip to the DB every time it is called to define the roles for the curren connection. `@WithRoles` is used liberally within the codebase, and is often called multiple times within the same transaction.

This unnecessary frequent calling adds an overhead to processing new datapoints

For this following upload, `setRoles()` utilized 35.62% of cpu time

Screenshot from 2024-05-22 06-20-21

barreiro:
Fixes #1726 

Sets the `horreum.system` role by default on the connection and only set roles when needed.

Also, the roles are set with transaction scope to avoid having to reset them.

This avoid most calls to `setRoles` for non-authentication usage and reduces to a single call when authenticated, for most cases.


johnaohara:
Comparing the flamegraphs of bulk import.

Before PR, `setRoles()` utilized 22.32% of cpu time: 

Screenshot from 2024-07-03 08-32-23


After PR, `setRoles()` utilized 0.22% of cpu time: 

Screenshot from 2024-07-03 08-32-27

barreiro:
alright !!

lampajr (approved):
LGTM

PR #1813, opened 2024-06-21
[java-client] Fallback to basic authentiticaton when no OIDC token is…

barreiro:
… acquired

### DIFF
--- a/horreum-client/src/main/java/io/hyperfoil/tools/auth/KeycloakClientRequestFilter.java
+++ b/horreum-client/src/main/java/io/hyperfoil/tools/auth/KeycloakClientRequestFilter.java
@@ -1,8 +1,8 @@
 package io.hyperfoil.tools.auth;
 
 import io.hyperfoil.tools.CustomResteasyJackson2Provider;
-import org.apache.http.conn.HttpHostConnectException;
 import org.jboss.logging.Logger;
+import org.jboss.resteasy.client.jaxrs.internal.BasicAuthentication;
 import org.jboss.resteasy.client.jaxrs.internal.ResteasyClientBuilderImpl;
 import org.jboss.resteasy.plugins.providers.FormUrlEncodedProvider;
 import org.keycloak.admin.client.Keycloak;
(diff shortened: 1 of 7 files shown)


horreum-client/src/main/java/io/hyperfoil/tools/auth/KeycloakClientRequestFilter.java
```diff
@@ -50,24 +57,23 @@ public KeycloakClientRequestFilter(String keycloakBaseUrl,
...
-				LOG.warnf("Aborting the request with HTTP 500 error");
-				requestContext.abortWith(Response.status(500).build());
-			} else {
-				LOG.warnf("Aborting the request with HTTP 401 error");
-				requestContext.abortWith(Response.status(401).build());
+			if (showAuthMethod) {
+				LOG.infov("Using Basic authentication as OIDC server replied with {0}", ex.getMessage());
+				showAuthMethod = false;
 			}
+			basicAuthentication.filter(requestContext);
```
johnaohara, line 72: I am confused about this fallback here, in `HorreumClient` there is a check for keycloak being configured `if (keycloakConfig.url == null || keycloakConfig.url.isEmpty())`, and then either `BasicAuthentication` or `KeycloakClientRequestFilter` is registered with the client builder.   

Here, if there is an exception, it is reverting to basicAuth.  

If a user configures keycloak, and the authentication fails, shouldn't we just error here?

barreiro: these are slightly different. `keycloakConfig.url` may not exist if authentication is handled by Horreum (without a keycloak instance) and in that case it makes no sense to try OIDC authentication.  

if the OIDC is out of our control, then the client will try to authenticate with and fallback to basic after it gets the `400` error from the server. this way there are no changes in the way clients are configured. 

at the moment the clients use the same `api/config/keycloak` endpoint the UI uses, and because of that can't know if the OIDC server is under keycloak's control or not. we could also improve the output of that endpoint to inform the clients which auth method to use.

barreiro: scratch that last comment. the client may have to use more that one authentication scheme to authenticate to a single Horreum instance. it may have to authenticate as a user or as a machine account. one example of that is the integration test !! 

an alternative is to add a parameter to configure the client (to select between `machine-account` and `user-account`), having one of those methods as default. this can be a breaking change though.

barreiro: I don't get the question. The test currently uses the client with the `horreum.bootstrap` user account (OIDC) to setup a machine account (Basic) . Machine accounts always use Basic scheme, while user accounts can be either OIDC or Basic. 

We can use two or more test profiles, but the bottom line is that the client needs to support both schemes at the same time.

johnaohara: I think this is ok for now, we might want to revisit when we remove implement API keys and possibly remove machine accounts

johnaohara (approved):

PR #1859, opened 2024-07-03
Offload multi run upload processing to message queue: Fixes #1858

### LINKED ISSUE #1858: Datastores can return large numbers runs that are processing synchronously
**Describe the bug**

Some backend datastore integrations can return a large number of runs.

This can cause issues with importing a lot of run data in one query, such as long transaction times, failure to process a single run data will fail the entire import etc.

The processing of run data should be performed asynchronously in these cases and runs should be offloaded to an external queue. 

See https://github.com/Hyperfoil/Horreum/pull/1703 for an example datastore that can return a large number of run results

johnaohara:
Offload processing multiple runs from datastore to async queue

## Fixes Issue

Fixes #1858

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
@@ -1,19 +1,11 @@
 package io.hyperfoil.tools.horreum.svc;
 
 import static com.fasterxml.jackson.databind.node.JsonNodeFactory.instance;
-import static io.hyperfoil.tools.horreum.entity.data.SchemaDAO.QUERY_1ST_LEVEL_BY_RUNID_TRANSFORMERID_SCHEMA_ID;
-import static io.hyperfoil.tools.horreum.entity.data.SchemaDAO.QUERY_2ND_LEVEL_BY_RUNID_TRANSFORMERID_SCHEMA_ID;
-import static io.hyperfoil.tools.horreum.entity.data.SchemaDAO.QUERY_TRANSFORMER_TARGETS;
+import static io.hyperfoil.tools.horreum.entity.data.SchemaDAO.*;
 
 import java.io.IOException;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.TreeMap;
+import java.util.*;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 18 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
```diff
@@ -570,17 +574,53 @@ Response addRunFromData(String start, String stop, String test,
...
       }
-//      return Response.status(Response.Status.OK).entity(String.valueOf(runId)).header(HttpHeaders.LOCATION, "/run/" + runId).build();
-      String reponseString = String.valueOf(runIds.stream().map(val -> Integer.toString(val)).collect(Collectors.joining(", ")));
-      return Response.status(Response.Status.OK).entity(reponseString).build();
+
+      responseString = runIds.size() > 0 ?
+              String.valueOf(runIds.stream().map(val -> Integer.toString(val)).collect(Collectors.joining(", ")))
+              : "More than 10 runs uploaded, processing asynchronously";
+
+      return Response.status(Response.Status.OK).entity(responseString).build();
```
lampajr, line 599: Can we return different status codes in according to whether the process has been sync or async?
E.g., `200` is the result contains the list of runIds and `202` is the results does not contain anything useful as it is going to be processed async, wdyt?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
```diff
@@ -570,17 +574,53 @@ Response addRunFromData(String start, String stop, String test,
...
    }
+   @Transactional
+   public void persistRun(ServiceMediator.RunUpload runUpload) {
```
lampajr, line 602: Otherwise we would expose `ServiceMediator.RunUpload` outside its scope.
```suggestion
   void persistRun(ServiceMediator.RunUpload runUpload) {
```

horreum-backend/src/main/resources/application.properties
```diff
@@ -58,6 +58,18 @@ mp.messaging.outgoing.schema-sync-out.address=schema-sync
 mp.messaging.outgoing.schema-sync-out.durable=true
 mp.messaging.outgoing.schema-sync-out.container-id=horreum-broker
 mp.messaging.outgoing.schema-sync-out.link-name=schema-sync
+# schema-sync incoming
```
lampajr, line 61:
```suggestion
# run-upload incoming
```

horreum-backend/src/main/resources/application.properties
```diff
@@ -58,6 +58,18 @@ mp.messaging.outgoing.schema-sync-out.address=schema-sync
...
+mp.messaging.incoming.run-upload-in.connector=smallrye-amqp
+mp.messaging.incoming.run-upload-in.address=run-upload
+mp.messaging.incoming.run-upload-in.durable=true
+mp.messaging.incoming.run-upload-in.container-id=horreum-broker
+mp.messaging.incoming.run-upload-in.link-name=run-upload
+# schema-sync outgoing
```
lampajr, line 67:
```suggestion
# run-upload outgoing
```

lampajr (commented):
Hi @johnaohara, not sure if the PR was really ready for review :smile: 
But I just left a couple of comments, mostly minors btw

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
johnaohara: Yes, 202 makes sense in this case.  I will change the return code and update the test

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
johnaohara: Hmm... `ServiceMediator.RunUpload` is already outside of scope here, because this method couples RunServiceImpl with ServiceMediator... which I don't think is a great pattern.  

tbh, i don't know how much of the codebase is tightly coupled like this

lampajr: >ServiceMediator.RunUpload is already outside of scope here

Well, from java "Access modifier" point of view if we remove the `public` from `persitRun` method it shouldn't be out of scope here but yeah I agree that this is going to thigthly couple the RunServiceImpl with ServiceMediator even if I don't think this is an issue here

lampajr (approved):
lgtm, thanks John!

PR #1880, opened 2024-07-09
Dynamically resolve custom views in Compare Datasets: Fixes #1879

### LINKED ISSUE #1879: Compare Datasets does not display defined Views correctly
**Describe the bug**

For a test that has custom views defined, the "Compare Datasets" function does not display the Views as tab, and the "Default" tab resolves randomly to one of the custom Views

Screenshot from 2024-07-09 16-49-11

Screenshot from 2024-07-09 16-49-22

johnaohara:
## Fixes Issue

Fixes #1879

Dynamically resolve the custom views for Compare Datasets

Screenshot from 2024-07-09 16-54-02

Screenshot from 2024-07-09 16-54-33

Screenshot from 2024-07-09 16-54-38

### DIFF
--- a/horreum-web/src/domain/runs/DatasetComparison.tsx
+++ b/horreum-web/src/domain/runs/DatasetComparison.tsx
@@ -85,7 +85,18 @@ export default function DatasetComparison() {
         [datasets]
     )
 
-    const defaultView = views?.find(v => (v.name = "Default"))
+    const fragmentTags = views.map(view => {
+        return (
+            <FragmentTab title={view.name} fragment={`view_${view.id}`} key={`view_${view.id}`}>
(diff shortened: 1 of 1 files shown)


lampajr (approved):
LGTM!

@johnaohara I just noticed that the labels comparison is actually displayed as separated page, maybe we could think to move the same under the specific test as we did for all the others. 

wdyt? if you agree I can open an issue

johnaohara:
> LGTM!
> 
> @johnaohara I just noticed that the labels comparison is actually displayed as separated page, maybe we could think to move the same under the specific test as we did for all the others.
> 
> wdyt? if you agree I can open an issue

@lampajr  Yes, the current implementation loses the navigation and breadcrumb. I think we need to bring this inline with the other views

PR #1894, opened 2024-07-16
Replace Test.get calls when not really required

lampajr:
## Changes proposed

- [x] Create ad-hoc method to simply check test existence and access right by triggering db level RLS
- [x] Replace `TestServiceImpl.get` call with the new one when not really required


lampajr:
**Benchmark config:**
- Calling `labelValues` endpoint
- Duration=360s
- Rate=20req/s

**Before the change:**
`TestServiceImpl.get` was using the `19.90%` of the cpu (~359 samples) and that was just used to check the test existence and access rights but without actually making use of the returned object. 
image

**After the change:**
The cpu usage of the test check has been reduced to `6.18%`.
image

lampajr:
Hey @johnaohara I preferred to use a plain query rather than `TestDao.count("id = ?1", testId)` but we can easily change it.

I'd say that this change is pretty much isolated so it should be safe, therefore rather than creating a dev branch with this included I think we can directly merge this in `master`.

> We can create the dev branch anyway to include any other improvement we might see

PR #1899, opened 2024-07-17
Improve getFilterDef by skipping unnecessary branches

lampajr:
## Changes proposed

- Improve the `getFilterDef` call by skipping some useless computation especially when the filter is blank
- Use StringBuilder to build filterSql, this should be helpful especially as we were concatenating String inside loops

## Check List (Check all the applicable boxes) 

- [ ] My code follows the code style of this project.
- [ ] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.

PR #1913, opened 2024-07-24
Improve label values

lampajr:
## Changes proposed

- [x] Based on https://github.com/Hyperfoil/Horreum/pull/1899
- [x] Remove `LIMIT` when not provided, i.e., get all
- [x] Remove `DISTINCT` from main query
- [x] Add index on `dataset.testId` and partial indexes on `label.filtering` and `label.metrics`

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
@@ -7,7 +7,6 @@ import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.JsonNodeType;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.hyperfoil.tools.horreum.api.SortDirection;
-import io.hyperfoil.tools.horreum.api.alerting.Change;
 import io.hyperfoil.tools.horreum.api.data.Access;
 import io.hyperfoil.tools.horreum.api.data.ExportedLabelValues;
 import io.hyperfoil.tools.horreum.api.data.Fingerprints;
@@ -42,12 +41,10 @@ import jakarta.annotation.security.RolesAllowed;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.persistence.EntityManager;
-import jakarta.persistence.NoResultException;
 import jakarta.persistence.PersistenceException;
 import jakarta.persistence.Tuple;
 import jakarta.transaction.TransactionManager;
 import jakarta.transaction.Transactional;
-import jakarta.ws.rs.DefaultValue;
 import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.Response;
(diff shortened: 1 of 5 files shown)


lampajr:
Postgres query plan running `/api/test/339/labelValues` without any query parameter and benchmark results with 20rps.

**Before**
```sql
 Limit  (cost=15670.89..15671.05 rows=64 width=56)
   ->  Sort  (cost=15670.73..15672.33 rows=637 width=56)
         Sort Key: dataset.runid DESC
         ->  Unique  (cost=15631.51..15641.07 rows=637 width=56)
               ->  Sort  (cost=15631.51..15633.10 rows=637 width=56)
                     Sort Key: (COALESCE(jsonb_object_agg(label.name, lv.value) FILTER (WHERE ((label.name IS NOT NULL) AND (label.name <> $1))), '{}'::jsonb)), dataset.runid, dataset.id, dataset.start, dataset.stop
                     ->  GroupAggregate  (cost=15123.68..15601.84 rows=637 width=56)
                           Group Key: dataset.id
                           ->  Gather Merge  (cost=15123.68..15565.43 rows=3793 width=102)
                                 Workers Planned: 2
                                 ->  Sort  (cost=14123.65..14127.60 rows=1580 width=102)
                                       Sort Key: dataset.id
                                       ->  Hash Left Join  (cost=294.29..14039.71 rows=1580 width=102)
                                             Hash Cond: (lv.label_id = label.id)
                                             Filter: ((label.id IS NULL) OR ($3 AND label.filtering) OR ($4 AND label.metrics))
                                             ->  Nested Loop Left Join  (cost=0.42..13738.01 rows=2981 width=55)
                                                   ->  Parallel Seq Scan on dataset  (cost=0.00..9179.24 rows=265 width=24)
                                                         Filter: (testid = $2)
                                                   ->  Index Scan using by_dataset on label_values lv  (cost=0.42..16.80 rows=40 width=35)
                                                         Index Cond: (dataset_id = dataset.id)
                                             ->  Hash  (cost=261.16..261.16 rows=2616 width=57)
                                                   ->  Seq Scan on label  (cost=0.00..261.16 rows=2616 width=57)
```
image


**After**
```sql
 Sort  (cost=13088.86..13090.42 rows=623 width=56)
   Sort Key: dataset.runid DESC
   ->  HashAggregate  (cost=13052.16..13059.95 rows=623 width=56)
         Group Key: dataset.id
         ->  Hash Left Join  (cost=296.31..13023.71 rows=3793 width=102)
               Hash Cond: (lv.label_id = label.id)
               Filter: ((label.id IS NULL) OR ($3 AND label.filtering) OR ($4 AND label.metrics))
               ->  Nested Loop Left Join  (cost=0.72..12709.30 rows=7155 width=55)
                     ->  Index Scan using dataset_testid_key on dataset  (cost=0.29..1856.27 rows=623 width=24)
                           Index Cond: (testid = $2)
                     ->  Index Scan using by_dataset on label_values lv  (cost=0.42..17.02 rows=40 width=35)
                           Index Cond: (dataset_id = dataset.id)
               ->  Hash  (cost=261.93..261.93 rows=2693 width=57)
                     ->  Seq Scan on label  (cost=0.00..261.93 rows=2693 width=57)
```
image

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -93,18 +90,18 @@ public class TestServiceImpl implements TestService {
...
    protected static final String LABEL_VALUES_QUERY = """
          WITH
          combined as (
-         SELECT DISTINCT COALESCE(jsonb_object_agg(label.name, lv.value) FILTER (WHERE label.name IS NOT NULL INCLUDE_EXCLUDE_PLACEHOLDER), '{}'::jsonb) AS values, runId, dataset.id AS datasetId, dataset.start AS start, dataset.stop AS stop
+         SELECT COALESCE(jsonb_object_agg(label.name, lv.value) FILTER (WHERE label.name IS NOT NULL INCLUDE_EXCLUDE_PLACEHOLDER), '{}'::jsonb) AS values, runId, dataset.id AS datasetId, dataset.start AS start, dataset.stop AS stop
                   FROM dataset
                   LEFT JOIN label_values lv ON dataset.id = lv.dataset_id
                   LEFT JOIN label ON label.id = lv.label_id
                   WHERE dataset.testid = :testId
                      AND (label.id IS NULL OR (:filteringLabels AND label.filtering) OR (:metricLabels AND label.metrics))
```
lampajr, line 102: Is there a reason why we need to keep records with `label.id = NULL`? IIUC this is only required to ensure that runs having all records with `label.id NULL` are associated with an empty `{}` jsonb.

Can't we just ignore them? and return all records where there is at least one label, properly associated?

lampajr:
Hi @willr3 @johnaohara, in order to properly compare the composable PoC with the current Horreum I tried to investigate why we have poor performance in the current setup.

At the moment this is what I was able to find, not huge improvement though!

I think that further improvements can be done if we re-design the query, e.g., using multiple queries rather than one single one or de-normalizing the data to avoid some joins or .. (?). Then, the same logic/reasoning could be then applied to the PoC as well. wdyt?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
willr3: This is the part that concerns me because it is a change to the result object for `/labelValues`. We have historically bumped the minor version whenever we made an API change. I'm also not sure what users would expect from current Horreum in terms of a payload for a dataset that doesn't have valid labels extracted. Should it be missing or included without values?

lampajr: yeah that's a good point, which is more from a semantic perspective.

btw this change is not included in this PR, so maybe that's something we could discuss separately as this could be also taken into account for the PoC.

willr3: I see the change to remove `DISTINCT` in the PR's diff for `TestServiceImpl` so I think it was pushed to the PR as well

lampajr: ah yeah sorry, I meant the removal of the `label.id IS NULL` is not part of this PR, the DISTINCT removal instead is here but AFAICS it did not change the results.. it should be backward compatible

willr3: If if doesn't change the results them I'm fine merging

Issue #1637, opened 2024-04-26
Code Formatter


lampajr:
**Feature idea.**

At the moment there isn't any automated/integrated code formatter, this could lead to some formatting issues where each dev applies his own formatter logic, perhaps based on the IDE (e.g., for the imports)

**Describe the solution you'd like**

Setup an automated and integrated formatter, even better if the configuration is not IDE-dependant.

We could setup a similar approach to what has been setup for [Hyperfoil](https://github.com/Hyperfoil/Hyperfoil/tree/master/config). 

**Additional information**

n/a

PR #1915, opened 2024-07-25
Enforce java codestyle

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/1637

## Changes proposed

- [x] Automate/integrate code formatting in the Maven lifecyle
- [x] Apply formatting on the current codebase

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/BackendResolver.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/BackendResolver.java
@@ -1,11 +1,12 @@
 package io.hyperfoil.tools.horreum.datastore;
 
-import io.hyperfoil.tools.horreum.api.data.datastore.DatastoreType;
-import io.quarkus.arc.All;
+import java.util.List;
+
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 
-import java.util.List;
+import io.hyperfoil.tools.horreum.api.data.datastore.DatastoreType;
+import io.quarkus.arc.All;
 
 @ApplicationScoped
 public class BackendResolver {
@@ -13,12 +14,11 @@ public class BackendResolver {
     @All
     List<Datastore> backendStores;
 
-    public Datastore getBackend(DatastoreType type){
+    public Datastore getBackend(DatastoreType type) {
         return backendStores.stream()
-                .filter( store -> store.type().equals(type))
+                .filter(store -> store.type().equals(type))
                 .findFirst()
                 .orElseThrow(() -> new IllegalStateException("Unknown datastore type: " + type));
     }
-
 
 }
--- a/horreum-client/src/main/java/io/hyperfoil/tools/HorreumClient.java
+++ b/horreum-client/src/main/java/io/hyperfoil/tools/HorreumClient.java
@@ -1,31 +1,6 @@
 package io.hyperfoil.tools;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import io.hyperfoil.tools.auth.KeycloakClientRequestFilter;
-import io.hyperfoil.tools.horreum.api.client.RunService;
-import io.hyperfoil.tools.horreum.api.internal.services.ActionService;
-import io.hyperfoil.tools.horreum.api.internal.services.AlertingService;
-import io.hyperfoil.tools.horreum.api.internal.services.BannerService;
-import io.hyperfoil.tools.horreum.api.internal.services.ChangesService;
-import io.hyperfoil.tools.horreum.api.services.ConfigService;
-import io.hyperfoil.tools.horreum.api.services.DatasetService;
-import io.hyperfoil.tools.horreum.api.services.ExperimentService;
-import io.hyperfoil.tools.horreum.api.internal.services.NotificationService;
-import io.hyperfoil.tools.horreum.api.internal.services.ReportService;
-import io.hyperfoil.tools.horreum.api.services.SchemaService;
-import io.hyperfoil.tools.horreum.api.internal.services.SqlService;
-import io.hyperfoil.tools.horreum.api.internal.services.SubscriptionService;
-import io.hyperfoil.tools.horreum.api.services.TestService;
-import io.hyperfoil.tools.horreum.api.internal.services.UserService;
-import org.jboss.resteasy.client.jaxrs.ResteasyClient;
-import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
-import org.jboss.resteasy.client.jaxrs.internal.BasicAuthentication;
-import org.jboss.resteasy.client.jaxrs.internal.ResteasyClientBuilderImpl;
-import org.jboss.resteasy.plugins.providers.DefaultTextPlain;
-import org.jboss.resteasy.plugins.providers.StringTextStar;
-
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.TrustManagerFactory;
+import static io.hyperfoil.tools.horreum.api.services.ConfigService.KEYCLOAK_BOOTSTRAP_URL;
 
 import java.io.Closeable;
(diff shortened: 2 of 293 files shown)


lampajr:
This is just a proposal as I think that using a unified (enforced) formatter will reduce the noise of typos/extra spaces/different indentation and so on.

Obviously this is going to be a huge pull request as the whole codebase will be re-formatted.

As codestyle I started from the one used in [Hyperfoil](https://github.com/Hyperfoil/Hyperfoil/blob/master/config/codestyle.xml) and I generated the _corresponding_ ecplise-profile configuration. The main difference is that, here, I used `4` as intendation/tab size but given that at the moment there are files using either 4, 3 or 2 I think we will change many files regardless of the decision :smile:

johnaohara:
@lampajr thanks for looking at this, since code formatting was disabled, inconsistencies have crept into the code base.

I like the idea of this being independent of IDE and env.

With this PR do we need to rely on the developer to compile the code *before* checking in changes?  We might want to have a PR hook that confirms that codestyle has been applied to a PR.  We used to have a git hook to force codestyle before the commit, but it was troublesome accross multiple envs (i.e. it didn't work in windows iirc)

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/BackendResolver.java
```diff
@@ -13,11 +13,10 @@ public class BackendResolver {
...
 
-    public Datastore getBackend(DatastoreType type){
-        return backendStores.stream()
-                .filter( store -> store.type().equals(type))
-                .findFirst()
-                .orElseThrow(() -> new IllegalStateException("Unknown datastore type: " + type));
+    public Datastore getBackend(DatastoreType type) {
+        return backendStores.stream().filter(store -> store.type().equals(type)).findFirst().orElseThrow(
+                () -> new IllegalStateException("Unknown datastore type: " + type)
+        );
```
johnaohara, line 19: I think we should be careful to preserve readability of code, especially when processing streams.

My personal preference when chaining methods together, i.e. with stream processing, is to split each method call out onto separate lines. 
I know this is a personal preference, but i find it really help when skimming a method to be able to read what a stream processor is doing

lampajr:
> With this PR do we need to rely on the developer to compile the code *before* checking in changes?  We might want to have a PR hook that confirms that codestyle has been applied to a PR.  We used to have a git hook to force codestyle before the commit, but it was troublesome accross multiple envs (i.e. it didn't work in windows iirc) 

I'd say yes, but even if the user does not format the source, the PR checks should fail as the `mvn package` will trigger it and the job will fail because there are uncomitted changes.

So I think we can setup a pre-commit hook that runs `mvn formatter:format` just before the commit and this is gonna work in most cases. wdyt?

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/BackendResolver.java
lampajr: Yeah I agree, I missed this option (I have the same preference)

Btw this is what I'd like to discuss, trying to find a common agreement on the codestyle

horreum-client/src/main/java/io/hyperfoil/tools/HorreumClient.java
```diff
@@ -183,22 +181,23 @@ public HorreumClient build() throws IllegalStateException {
...
+                            DatasetService.class
+                    ).build(), target.proxyBuilder(ExperimentService.class).build(), target.proxyBuilder(
+                            NotificationService.class
+                    ).build(), target.proxyBuilder(ReportService.class).build(), new RunServiceExtension(
+                            target, target.proxyBuilder(RunService.class).build()
+                    ), target.proxyBuilder(SchemaService.class).build(), target.proxyBuilder(
+                            SqlService.class
+                    ).build(), target.proxyBuilder(SubscriptionService.class).build(), target.proxyBuilder(
+                            TestService.class
+                    ).build(), target.proxyBuilder(UserService.class).build()
```
johnaohara, line 199: I find this format strange, formatting alternate parameters differently

lampajr: +1, it is an issue but I think I found the proper option.. will change it in a while

lampajr: With the latest change, this is formatted as:
```java
            return new HorreumClient(
                    client,
                    target.proxyBuilder(ActionService.class)
                            .build(),
                    target.proxyBuilder(AlertingService.class)
                            .build(),
                    target.proxyBuilder(BannerService.class)
                            .build(),
                    target.proxyBuilder(ChangesService.class)
                            .build(),
                    target.proxyBuilder(ConfigService.class)
                            .build(),
                    target.proxyBuilder(DatasetService.class)
                            .build(),
                    target.proxyBuilder(ExperimentService.class)
                            .build(),
                    target.proxyBuilder(NotificationService.class)
                            .build(),
                    target.proxyBuilder(ReportService.class)
                            .build(),
                    new RunServiceExtension(
                            target,
                            target.proxyBuilder(RunService.class)
                                    .build()
                    ),
                    target.proxyBuilder(SchemaService.class)
                            .build(),
                    target.proxyBuilder(SqlService.class)
                            .build(),
                    target.proxyBuilder(SubscriptionService.class)
                            .build(),
                    target.proxyBuilder(TestService.class)
                            .build(),
                    target.proxyBuilder(UserService.class)
                            .build()
            );
```
which seems much better

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/BackendResolver.java
lampajr: Now it should follow what you were suggesting:
```java
    public Datastore getBackend(DatastoreType type) {
        return backendStores.stream()
                .filter(
                        store -> store.type()
                                .equals(type)
                )
                .findFirst()
                .orElseThrow(() -> new IllegalStateException("Unknown datastore type: " + type));
    }
```

config/pre-commit
lampajr, line 1: Added `pre-commit` hook that will be triggered just before commit to ensure everything is properly formatted (users need to copy this hook to their local `.git/hooks` folder)

lampajr:
Added imports sorting, following the same approach that Quarkus is already doing in its codebase.

lampajr:
At the moment this PR includes a custom eclipse formatting configuration but I think a good alternative (as we did in Hyperfoil) is to make use of the Quarkus formatting configuration (see https://github.com/quarkusio/quarkus/blob/b1de140564c7cf77b0eec986ea5dea5108a40027/independent-projects/ide-config/src/main/resources/eclipse-format.xml) - this way we use a well-known and used configuration and we don't create another one.

Wdyt?

lampajr:
> I think this would be useful to have! Do you mind adding a small doc section regarding how to set this up for developers?

yeah sure, I was thinking of adding something similar to what I've added in Hyperfoil in the `CONTRIBUTING` file https://github.com/Hyperfoil/Hyperfoil/blob/master/CONTRIBUTING.md.

Will update this PR asap :rocket:

lampajr:
> I think this would be useful to have! Do you mind adding a small doc section regarding how to set this up for developers?

Done, I have also:
- Rebased the PR
- Changed the formatting configuration to the quarkus one as discussed in the last meeting
- Removed the `pre-commit` hook as per our discussion (I added this option in the CONTRIBUTING file if anyone wants to adopt it)
- Changed the author of the formatting to `rh-appservices-perf`

fyi @stalep @johnaohara 

> NOTE: if you are interested in the configuration that has been added here, you can check commit https://github.com/Hyperfoil/Horreum/pull/1915/commits/a6e01855cc970305b8b6b16e369cff824f2ec73d only

johnaohara (approved):
lgtm

PR #1973, opened 2024-09-09
Return empty list if machine role is not present

lampajr:
At the moment, everytime a user open the administrator tab a warning saying that it is not able to list machine accounts for a specific team (the one selected by default, i.e., the first one) alongside the exception which says that the role is NOT FOUND is printed in the logs.

```bash
2024-09-09 10:50:46,107 e4b3085409dd quarkus-run.jar[7] WARN  [io.hyp.too.hor.svc.use.KeycloakUserBackend] (executor-thread-23917) Unable to list machine accounts for team <MY_TEAM>: jakarta.ws.rs.NotFoundException: HTTP 404 Not Found
	at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.handleErrorStatus(ClientInvocation.java:242)
	at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.extractResult(ClientInvocation.java:216)
	at org.jboss.resteasy.client.jaxrs.internal.proxy.extractors.BodyEntityExtractor.extractEntity(BodyEntityExtractor.java:59)
	at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientInvoker.invokeSync(ClientInvoker.java:136)
	at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientInvoker.invoke(ClientInvoker.java:103)
	at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientProxy.invoke(ClientProxy.java:102)
	at jdk.proxy2/jdk.proxy2.$Proxy121.getUserMembers(Unknown Source)
	at io.hyperfoil.tools.horreum.svc.user.KeycloakUserBackend.machineAccounts(KeycloakUserBackend.java:368)
	at io.hyperfoil.tools.horreum.svc.user.KeycloakUserBackend_ClientProxy.machineAccounts(Unknown Source)
	at io.hyperfoil.tools.horreum.svc.UserServiceImpl.safeMachineAccounts(UserServiceImpl.java:215)
	at io.hyperfoil.tools.horreum.svc.UserServiceImpl.teamMembers(UserServiceImpl.java:100)
```

Given that the situation when the machine role is not present is an expected condition I don't see real value in logging the exception alongside the warning, this is going to pollute the log files with unnecessary and useless information.

Therefore I'd propose to catch the `NotFoundException` and either log it at `DEBUG` level or keep it as `WARNING` but without the stacktrace.

## Changes proposed

Instead of logging the exception when the `machine` role is not found, simply log the warning and return empty list.
This will cleanup the logs a little bit by removing useless information from them.

With the change the log will be something like:
```bash
2024-09-09 14:50:08,047 WARN  [io.hyp.too.hor.svc.use.KeycloakUserBackend] (executor-thread-1) Unable to list machine accounts for team <MY_TEAM>
```

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
@@ -417,6 +417,9 @@ public class KeycloakUserBackend implements UserBackEnd {
             String prefix = getTeamPrefix(team);
             return keycloak.realm(realm).roles().get(prefix + Roles.MACHINE).getUserMembers(0, Integer.MAX_VALUE).stream()
                     .map(KeycloakUserBackend::toUserInfo).toList();
+        } catch (NotFoundException ex) {
+            LOG.debugv("Unable to list machine accounts for team {0}", team);
+            return List.of();
         } catch (Throwable t) {
             LOG.warnv(t, "Unable to list machine accounts for team {0}", team);
             throw ServiceException


johnaohara:
@lampajr is this `HTTP 404 Not Found` response because `org.keycloak.admin.client.resource.RoleResource.getUserMembers()` returns no members, or is it because the admin client is calling an incorrect URL?

I would have expected (or hoped) that if there are no members, that an empty list was returned

Do we have a mismatch between the client version and the keycloak version?

johnaohara:
In the 0.14 branch, we depend on the keycloak 24.0.4 client.  In our env we are running keycloak server 23.0.3

lampajr:
>is this HTTP 404 Not Found response because org.keycloak.admin.client.resource.RoleResource.getUserMembers() returns no members, or is it because the admin client is calling an incorrect URL?

@johnaohara  IIUC it is because it looks for a role like `TEAM-machine` and it is not configured in Keycloak for that team (which is the current scenario in prod)

lampajr:
> In the 0.14 branch, we depend on the keycloak 24.0.4 client. In our env we are running keycloak server 23.0.3

I hit the same in my local env where we use https://github.com/Hyperfoil/Horreum/blob/629c0fc7626c69c29d8ce01eeb9c5cffb201fba8/pom.xml#L97 for the Keycloak dev-service so I don't think it a compatibility issue but just a not found role

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
```diff
@@ -417,6 +417,9 @@ public List<UserService.UserData> machineAccounts(String team) {
             String prefix = getTeamPrefix(team);
             return keycloak.realm(realm).roles().get(prefix + Roles.MACHINE).getUserMembers(0, Integer.MAX_VALUE).stream()
                     .map(KeycloakUserBackend::toUserInfo).toList();
+        } catch (NotFoundException ex) {
+            LOG.warnv("Unable to list machine accounts for team {0}", team);
```
johnaohara, line 421: If this is an expected scenario, i.e. if a team does not have machine accounts, then I think the asppropriate log level should be `debug`

johnaohara (changes requested):

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/user/KeycloakUserBackend.java
lampajr: >Therefore I'd propose to catch the NotFoundException and either log it at DEBUG level or keep it as WARNING but without the stacktrace.

yeah as I mentioned in the description, that was exactly my doubt.. whether debug or warning.
I agree that given it is an expected scenario, DEBUG is much more suitable here

lampajr: Updated

johnaohara (approved):
lgtm

PR #2002, opened 2024-09-17
Remove reactive datasource

barreiro:
I don't see the reactive database being used anywhere, and don't get test errors either. Therefore, I propose it's removal.


johnaohara (approved):
LGTM

PR #2025, opened 2024-09-24
Downgrade Keycloak to 23.0.7

johnaohara:
Keycloak version was incompatible with services running SSO

## Fixes Issue

Fixes: #2061

### DIFF
--- a/horreum-backend/pom.xml
+++ b/horreum-backend/pom.xml
@@ -150,6 +150,20 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-keycloak-admin-resteasy-client</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-admin-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-common</artifactId>
+        </dependency>
+
         <dependency>
             <groupId>io.quarkus</groupId>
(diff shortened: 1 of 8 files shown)


lampajr (changes requested):
Hey @johnaohara, I am not 100% sure but I think that we are still bringing `25.0.2` version for keycloak (in some places):

```

[INFO] \- io.quarkus:quarkus-keycloak-admin-resteasy-client:jar:3.14.4:compile
....
[INFO]    +- org.keycloak:keycloak-admin-client:jar:25.0.2:compile
```

That's might be the reason of the failure we see here.

johnaohara:
either we need to remove the dependency on `quarkus-keycloak-admin-resteasy-client` or only use Quarkus LTS releases

lampajr:
> it's a mess after we upgraded quarkus :(

+1

> either we need to remove the dependency on quarkus-keycloak-admin-resteasy-client 

I was trying this locally, but even if that works for the test I am not fully sure we won't introduce any bug by forcing  `quarkus-keycloak-admin-resteasy-client` to use an older version of keycloak

> only use Quarkus LTS releases

I think this will be something we should start considering..

lampajr:
> either we need to remove the dependency on `quarkus-keycloak-admin-resteasy-client` or only use Quarkus LTS releases

Both options are not so straightforward.. :sweat:

johnaohara:
@lampajr i saw this error when re-using a prod backup that had been loaded in dev mode by current main or 0.15 branch.  When I tested with a fresh backup it worked without error.

I think the with the current main branch, upgrading to keycloak 0.25.x will have modified the database tables, which are incompatible with 0.23.x, so downgrading keycloak on an already migrated database will break auth.

We need to be able to chose the keycloak version to maintain compatibility with production versions

lampajr (approved):
lgtm then!

PR #2035, opened 2024-09-30
Remove APIUtil class

barreiro:
As far as I can see, this is not in use anywhere. 

The JSON encoding/decoding properties are being set on `JavaTimeCustomizer` for the backend.


lampajr (approved):
lgtm

PR #2041, opened 2024-09-30
Add support for jsonpath filter expression ...

afcollins:
... which may return an ArrayNode that we need to process more

## Fixes Issue

When `findJsonPath` is invoked with a jsonpath filter expression, the ArrayNode is not transformed correctly.

## Changes proposed

If the result of the jsonPath is an ArrayNode of size exactly one, then process that one-and-only entry as a value.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
@@ -431,9 +431,9 @@ public class Util {
         try {
             JsonPath path = JsonPath.compile(jsonPath);
             Object obj = ctx.read(path);
-            if (obj instanceof ArrayNode) {
-                if (((ArrayNode) obj).size() == 1) {
-                    obj = ((ArrayNode) obj).get(0);
+            if (obj instanceof ArrayNode arr) {
+                if (arr.size() == 1) {
+                    obj = arr.get(0);
                 }
             }
             if (obj instanceof ValueNode) {
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
@@ -431,6 +431,11 @@ public class Util {
         try {
             JsonPath path = JsonPath.compile(jsonPath);
             Object obj = ctx.read(path);
+            if (obj instanceof ArrayNode) {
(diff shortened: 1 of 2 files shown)


horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/UtilTest.java
```diff
@@ -118,6 +121,41 @@ public void close() throws IOException {
...
+        ObjectNode cdv2Summary = JsonNodeFactory.instance.objectNode();
+        String cdv2End = "2024-09-05T18:14:42.797624736Z";
+        cdv2Summary.put("endTimestamp", cdv2End);
+        String cdv2Start = "2024-09-05T17:49:02.915410546Z";
+        cdv2Summary.put("timestamp", cdv2Start);
+        values.add(cdv2Summary);
+        ObjectNode cdv2JobConfig = cdv2Summary.putObject("jobConfig");
+        cdv2JobConfig.put("name", "cluster-density-v2");
+
+        System.out.println(root);
```
johnaohara, line 152: Please can you either remove this println, or replace it with a `Log.trace(root);` statement, so that we can control logging while running tests, thanks

afcollins: Removed

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/Util.java
```diff
@@ -431,6 +431,11 @@ public static Object findJsonPath(JsonNode input, String jsonPath) {
         try {
             JsonPath path = JsonPath.compile(jsonPath);
             Object obj = ctx.read(path);
+            if (obj instanceof ArrayNode) {
+                if (((ArrayNode) obj).size() == 1) {
+                    obj = ((ArrayNode) obj).get(0);
+                }
+            }
```
lampajr, line 438: This could be simplified with something like this:

```suggestion
            if (obj instanceof ArrayNode arr) {
                if (arr.size() == 1) {
                    obj = arr.get(0);
                }
            }
```

lampajr (approved):
Thanks @afcollins  looks good to me, I left just a minor comment.

As this is just a simplification I'd let up to you to decide whether to update it or not.

afcollins:
Thanks! This pattern matching for instanceof is cleaner.

lampajr (approved):

Issue #2078, opened 2024-10-10
Horreum labelValues before and after filtering is not working


lampajr:
**Describe the bug**

Looks like that `before` and `after` filtering on `labelValues` endpoint is not working as expected, no results are returned even when there should be.

**To Reproduce**

Call `labelValues` with either `before` or `after` or both.

```bash
curl 'https://horreum.corp.redhat.com/api/test/TEST_ID/labelValues?after="2024-10-07T20:20:32.183Z"'
```

Replace TEST_ID with your specific test id

**Version**

What is the version of Horreum ?

If you are using a development branch; what is the commit id ?

```
0.15.x
```

lampajr:
Adding `priority/hold` as it looks like the issue that was raised is not reproducible and the logic seems working as expected.

lampajr:
This has not been confirmed as an issue, the filerting looks working as expected, therefore closing this issue as "won't do".

With https://github.com/Hyperfoil/Horreum/pull/2081 I added a couple of more use cases that improved a little bit the coverage.

PR #2080, opened 2024-10-11
Fix python and hunter version

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2079

## Changes proposed

- [x] Fix the python and hunter version in the CI
- [x] Fix the Hunter version in the Horreum container

## Check List (Check all the applicable boxes) 

- [x] My code follows the code style of this project.
- [ ] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.

PR #2081, opened 2024-10-14
Add before/after labelValues filtering test

lampajr:
## Fixes Issue

This was initially meant to address https://github.com/Hyperfoil/Horreum/issues/2078 but it looks like the before/after filtering is actually working as expected.

Therefore, at the moment, I am proposing to simply add some more testing that will improve the labelValues coverage.

## Changes proposed

- [x] Add some more testing on before/after labelValues filtering


lampajr:
Current failure during the Hunter installation should be fixed by https://github.com/Hyperfoil/Horreum/pull/2080.

PR #2083, opened 2024-10-14
Remove machine accounts

barreiro:
The introduction of API keys (#1952) as an alternative autherntication method to machine accounts makes that feature obsolete.

Let's remove the machine accounts right away as that authentication method that is not needed anymore.


lampajr (approved):
LGTM!

Issue #1975, opened 2024-09-11
Recalculation process blocked waiting for lock release


lampajr:
**Describe the bug**

Looks like that some processes are blocked waiting for {{lock}} to be released at https://github.com/Hyperfoil/Horreum/blob/69fdfe81cfa2e1aa4d839c2311bc4f03f0d41bc1/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java#L565-L572

This could block the thread for long time, especially considering that the recalculation could take long time:

```bash
2024-09-09 16:08:13,718 797ee630d871 quarkus-run.jar[7] WARN  [io.ver.cor.imp.BlockedThreadChecker] (vertx-blocked-thread-checker) Thread Thread[horreum.run.pool-2,5,main] has been blocked for 61418 ms, time limit is 60000 ms: io.vertx.core.VertxException: Thread blocked
	at java.base@17.0.11/jdk.internal.misc.Unsafe.park(Native Method)
	at java.base@17.0.11/java.util.concurrent.locks.LockSupport.park(LockSupport.java:211)
	at java.base@17.0.11/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:715)
	at java.base@17.0.11/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:938)
	at java.base@17.0.11/java.util.concurrent.locks.ReentrantLock$Sync.lock(ReentrantLock.java:153)
	at java.base@17.0.11/java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:322)
	at io.hyperfoil.tools.horreum.svc.DatasetServiceImpl.withRecalculationLock(DatasetServiceImpl.java:554)
	at io.hyperfoil.tools.horreum.svc.DatasetServiceImpl.onNewDataset(DatasetServiceImpl.java:563)
	at io.hyperfoil.tools.horreum.svc.DatasetServiceImpl_ClientProxy.onNewDataset(Unknown Source)
	at io.hyperfoil.tools.horreum.svc.ServiceMediator.newDataset(ServiceMediator.java:141)
	at io.hyperfoil.tools.horreum.svc.ServiceMediator_ClientProxy.newDataset(Unknown Source)
	at io.hyperfoil.tools.horreum.svc.RunServiceImpl.createDataset(RunServiceImpl.java:1355)
[...]

PR #2084, opened 2024-10-14
Do not use recalc lock for run-recalc

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/1975

## Changes proposed

- [x] Do not use shared lock during labelValues computation 
- [x] Move shared lock usage into ServiceMediator if really required 
- [x] Manage concurrent runs transformation similarly to what we are doing for test datasets recalculation
- [x] Fix the test datasets recalculation modal (show correct data)
   
image

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
@@ -64,6 +65,7 @@ import io.hyperfoil.tools.horreum.api.data.Run;
 import io.hyperfoil.tools.horreum.api.data.ValidationError;
 import io.hyperfoil.tools.horreum.api.services.RunService;
 import io.hyperfoil.tools.horreum.api.services.SchemaService;
+import io.hyperfoil.tools.horreum.api.services.TestService;
 import io.hyperfoil.tools.horreum.bus.AsyncEventChannels;
 import io.hyperfoil.tools.horreum.datastore.BackendResolver;
 import io.hyperfoil.tools.horreum.datastore.Datastore;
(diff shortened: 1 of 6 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -720,7 +720,7 @@ public void recalculateDatasets(int testId) {
             // transform will add proper roles anyway
             //         messageBus.executeForTest(testId, () -> datasetService.withRecalculationLock(() -> {
             //         mediator.executeBlocking(() -> mediator.transform(runId, true));
-            mediator.executeBlocking(() -> mediator.withRecalculationLock(() -> {
+            mediator.executeBlocking(() -> mediator.withSharedLock(() -> {
```
lampajr, line 723: Not sure if we can remove the lock usage also here 🤔

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
```diff
@@ -1118,14 +1122,38 @@ public void recalculateAll(String fromStr, String toStr) {
...
+
         log.debugf("Transforming run ID %d, recalculation? %s", runId, Boolean.toString(isRecalculation));
+        int numDatasets = 0;
+
+        // check whether there is an ongoing transformation on the same runId
+        TestService.RecalculationStatus status = new TestService.RecalculationStatus(1);
+        TestService.RecalculationStatus prev = transformations.putIfAbsent(runId, status);
+        // ensure the transformation is removed, with this approach we should be sure
+        // it gets removed even if transaction-level exception occurs, e.g., timeout
+        Util.registerTxSynchronization(tm, txStatus -> transformations.remove(runId, status));
```
lampajr, line 1150: I couldn't use `remove(runId, null)` as the `remove` methods expect a non-null value to properly remove the object:

```java
   public boolean remove(Object key, Object value) {
        if (key == null)
            throw new NullPointerException();
        return value != null && replaceNode(key, null, value) != null;
    }
```

lampajr (commented):
@johnaohara as discussed offline, I updated the implementation such that:

- There is no timeout
- The transformation gets remove from the map as transaction callback
- If there exists a transformation for a given runId it will simply return 0

johnaohara (approved):

PR #2100, opened 2024-10-18
Remove test token feature

barreiro:
The introduction of API keys (https://github.com/Hyperfoil/Horreum/pull/1952) as an alternative authentication mechanism makes the test tokens redundant.

We're now at a point when that feature can be removed.


horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/RunService.java
```diff
@@ -46,30 +57,25 @@ public interface RunService {
     @Operation(description = "Get extended Run information by Run ID")
     @Parameters(value = {
             @Parameter(name = "id", in = ParameterIn.PATH, description = "Run ID", example = "202"),
-            @Parameter(name = "token", in = ParameterIn.QUERY, description = "Run API token", example = "094678029a2aaf9a2847502273099bb3a1b2338c2b9c618ed09aef0181666e38")
 
     })
-    RunExtended getRun(@PathParam("id") int id,
-            @QueryParam("token") String token);
+    RunExtended getRun(@PathParam("id") int id);
```
barreiro, line 62: should the endpoints that take a token be kept (deprecated) for compatibility, sending an error when a token parameter exists ?!? ... or is it fine to just drop the token altogether ??
 
the same applies to `getRunSummary`, `getMetaData`, etc.

lampajr: IIRC query params are backward compatible in the sense that users can still provide them but they won't actually work (i.e., will be ignored).

On the other this change might affect existing users that are relying on this token feature, so we should clearly state that this has been removed and we should guide them through the new API token feature (i.e., a sort of migration guide from old tokens to new API tokens).

To answer your question, IMO I don't see a real value in keeping the query param just to return an error.. I think it is much better to completely remove it and add new migration guide to help existing users.

johnaohara: I think it is fine to handle this in a migration guide and drop support in the API.  This will break integrations, but as long as the braking change is documented, with clear guidance on how to migrate, I think that is fine

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/TestService.java
johnaohara, line 1: We still have a token in the api at;

```
    @GET
    @Path("{id}")
    @Operation(description = "Retrieve a test by id")
    Test get(@PathParam("id") int id, @QueryParam("token") String token);
```

There is also an erroneous mention of a token at `@Parameter(name = "id", required = true, description = "Test ID to revoke **token**", example = "101")`;

```
    @POST
    @Path("{id}/updateAccess")
    @Operation(description = "Update the Access configuration for a Test")
    @Parameters(value = {
            @Parameter(name = "id", required = true, description = "Test ID to revoke **token**", example = "101"),
            @Parameter(name = "owner", required = true, description = "Name of the new owner", example = "perf-team"),
            @Parameter(name = "access", required = true, description = "New Access level for the Test", example = "0")
    })
    void updateAccess(@PathParam("id") int id,
            @QueryParam("owner") String owner,
@QueryParam("access") Access access);
```

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/EncryptionManager.java
```diff
@@ -1,66 +0,0 @@
...
-import javax.crypto.spec.SecretKeySpec;
-
-import jakarta.annotation.PostConstruct;
-import jakarta.enterprise.context.ApplicationScoped;
-
-import org.eclipse.microprofile.config.inject.ConfigProperty;
-
-@ApplicationScoped
-public class EncryptionManager {
-    @ConfigProperty(name = "horreum.db.secret")
```
johnaohara, line 25: Can we finally remove `horreum.db.secret` everywhere? iirc, tokens were the last part of the system still using this property. 

Although it looks like the DB table has been dropped for a while, we can remove horreum.db.secret everywhere is it mentioned in docs, and scripts

horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -4598,4 +4598,70 @@
...
+
+            ALTER POLICY action_policy ON action USING (has_role('horreum.system') OR has_role('admin') OR has_role2((SELECT owner FROM test WHERE test.id = test_id), 'tester'));
+
+            ALTER POLICY dataset_insert ON dataset WITH CHECK (has_role2(owner, 'uploader') OR has_role('horreum.system'));
+
+            DROP FUNCTION can_view; <!-- calls to this function are replaced with existing `can_view2` -->
+            DROP FUNCTION has_read_token;
+            DROP FUNCTION has_modify_token;
+            DROP FUNCTION has_upload_token;
+        </sql>
```
johnaohara, line 4662: I think we can also safely drop the `auth_suffix` function

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
```diff
@@ -374,10 +342,9 @@ public void updateAccess(int id, String owner, Access access) {
 
     @PermitAll // all because of possible token-based upload
```
johnaohara, line 343: I think we need to restrict this to `@RolesAllowed(Roles.UPLOADER)` to limit the to users with that role

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/RunServiceImpl.java
```diff
@@ -458,16 +423,14 @@ public Response addRunFromData(String start, String stop, String test, String ow
             log.error("Failed to read data/metadata from upload file", e);
             throw ServiceException.badRequest("Provided data/metadata can't be read (JSON encoding problem?)");
         }
-        return addRunFromData(start, stop, test, owner, access, token, schemaUri, description, dataNode.toString(),
-                metadataNode);
+        return addRunFromData(start, stop, test, owner, access, schemaUri, description, dataNode.toString(), metadataNode);
     }
 
     @PermitAll // all because of possible token-based upload
```
johnaohara, line 429: I think we need to restrict this to `@RolesAllowed(Roles.UPLOADER)` to limit the to users with that role

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -131,7 +125,6 @@ public void delete(int id) {
     }
 
     @Override
-    @WithToken
     @WithRoles
     @PermitAll
     public Test get(int id, String token) {
```
johnaohara, line 130: Need to remove `token` from this method signature

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -174,7 +167,7 @@ protected boolean checkTestExists(int id) {
     }
 
     @WithRoles(extras = Roles.HORREUM_SYSTEM)
```
johnaohara, line 169: I think this is no longer required and we can remove it. Elevated privileges here is a security risk.

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/BaseServiceTest.java
```diff
@@ -192,12 +193,13 @@ protected void dropAllViewsAndTests() {
 
                 em.createNativeQuery("DELETE FROM test_transformers").executeUpdate();
                 em.createNativeQuery("DELETE FROM transformer_extractors").executeUpdate();
+                em.createNativeQuery("DELETE FROM experiment_comparisons").executeUpdate();
```
johnaohara, line 196: Does this relate to another issue?

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/RunServiceTest.java
```diff
@@ -756,27 +756,6 @@ public void testUploadToPrivateTest() throws JsonProcessingException {
...
-        RestAssured.given()
-                .header(HttpHeaders.CONTENT_TYPE, "application/json")
-                .body(org.testcontainers.shaded.com.fasterxml.jackson.databind.node.JsonNodeFactory.instance.objectNode())
-                .post("/api/run/data?start=" + now + "&stop=" + now + "&test=" + test.name + "&owner=" + UPLOADER_ROLES[0] +
-                        "&access=" + Access.PRIVATE + "&token=" + MY_SECRET_TOKEN)
-                .then()
-                .statusCode(200)
-                .extract().asString();
-    }
-
```
johnaohara, line 779: Do we have an equivalent test that uploads to a private test using an API key?

johnaohara (changes requested):
Please see comments

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/server/EncryptionManager.java
lampajr: +1 I think that could have been removed already but I agree to remove all references in this PR directly

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/BaseServiceTest.java
barreiro: no, I get ` ERROR: update or delete on table "variable" violates foreign key constraint "experiment_comparison_variable" on table "experiment_comparisons"
  Detail: Key is still referenced from table "experiment_comparisons".` errors without it.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
barreiro: I get errors when I remove it so let's leave it for now.

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/RunServiceTest.java
barreiro: I have added an integration test.

johnaohara (approved):

johnaohara:
> Overall looks good to me but I'd update the documentation accordingly, i.e., removing any mention to the Horreum token like https://horreum.hyperfoil.io/docs/tasks/upload-run/#getting-horreum-token and adding a new migration guide (even very simple) clearly stating out to migrate from Horreum token to the API token approach.

I have opened #2154 to track the guide

PR #2107, opened 2024-10-21
Fix JSON deserialization of elasticsearch request type

### LINKED ISSUE #2097: Elastic Search import ENUM  type deserialization error for JSON payload.
**Describe the bug**

 Users that are providing a Payload in JSON provide a `type` attribute.
 Currently users defining the type in lower case will always see the request rejected.

Could not parse request: {"type":"search","index":"blah","query":{"from":0,"size":3,"query":{"bool":{"must":[{"term":{"name.keyword":"Foo bar baz"}}]}}, Cannot deserialize value of type `io.hyperfoil.tools.horreum.datastore.ElasticsearchDatastore$RequestType` from String "search": not one of the values accepted for Enum class: [SEARCH, MULTI_INDEX, DOC]}} 

**To Reproduce**

Provide an upload JSON with the type in lower or mixed case.

{"type":"search","index":"blah","query":{"from":0,"size":3,"query":{"bool":{"must":[{"term":{"name.keyword":"Foo bar baz"}}]}}

**Expected outcome**

The type is converted to upper case before attempting the ENUM lookup.

**Version**

0.15.6

**Java**

17

barreiro:
Fixes #2097

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/ElasticsearchDatastore.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/datastore/ElasticsearchDatastore.java
@@ -23,6 +23,7 @@ import org.elasticsearch.client.RestClient;
 import org.elasticsearch.client.RestClientBuilder;
 import org.jboss.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


lampajr (approved):
lgtm!

PR #2119, opened 2024-10-23
Add base roles to API keys

barreiro:
If a user has a composite role (like `dev-uploader`) it will get the base roles from keycloak (in this case `dev-team` and `uploader`) as well.

At the moment the API key only provides the composite roles. It should give the same set of composite and base roles.


lampajr (changes requested):
Hi @barreiro , I still need to finalize the review but meanwhile I do have some comments below:

Moreover,  it seems you missed a code formatting change (the "check uncommitted changes" chek failed)

lampajr (approved):
lgtm

PR #2128, opened 2024-10-24
fix delete labels from the ui

barreiro:
the definition of `deleteLabel` is 
``` javascript
deleteLabel(schemaId: number, labelId: number)
```


lampajr (approved):
lgtm

PR #2133, opened 2024-10-25
Improve response sstatus code when query return no results

barreiro:
although nothing is displayed in the UI, it may help to debug in situations where the query is valid but no runs are created.


lampajr (approved):
LGTM!

PR #2162, opened 2024-11-06
Datasets listing should treat {} as empty filter

lampajr:
## Changes proposed

Early improvement that would consider `{}` as empty filter, this way the query is a LOT easier and faster.

This does not solve the performance issue we currently have when the filter is set, as it would require a refactoring that might be not necessarily as we are trying to get rid of the datasets entirely.


johnaohara (approved):
LGTM

PR #2181, opened 2024-11-22
Fix Horreum and AMQ test containers

barreiro:
1. change `ARTEMIS_USER / ARTEMIS_PASSWORD` to `AMQ_USER / AMQ_PASSWORD` that are the variable names defined in https://github.com/artemiscloud/activemq-artemis-broker-image/blob/main/image.yaml 
2. add wait strategy to both AMQ and Horreum containers
3. misc code improvements


lampajr (approved):
looks working to me!

PR #2194, opened 2024-12-02
Fix `runApi.addRunFromData()` invocation

zakkak:
## Changes proposed

Removes redundant argument that used to be internal/test `token`.  The
parameter was removed in https://github.com/Hyperfoil/Horreum/pull/2100

This essentially fixes the "Import Data"/"Import New Run" form which currently always results in using the `schemaUrn` as a description, and `undefined` as the payload.


zakkak:
@lampajr @johnaohara this will need to be backported to 0.16

lampajr (approved):
Thanks a lot @zakkak, great catch!

Issue #2201, opened 2024-12-05
Horreum procedure `calc_dataset_view` computes wrong data


lampajr:
**Describe the bug**

 It turned out that currently the process of recomputing the label values is quite tricky and it involves some procedure and triggers to compute the `dataset_view` which is the one showed when filtering datasets.

```
calculateLabelValues -> insert "label_values" -> call calc_dataset_views

                                                -> triggers dsv_after_lv_insert_func -> insert "view_recalc_queue" -> triggers recalc_dataset_view
```
As you can see we are triggering both `calc_dataset_views` and `recalc_dataset_view` in an unpredictable order.

The problem is that `calc_dataset_views` is wrong and it can compute wrong data.
https://github.com/Hyperfoil/Horreum/blob/a763314753afe7c01890146f74a1f3667c5c221a/horreum-backend/src/main/resources/db/changeLog.xml#L4321-L4334

It is joining `label_values` without adding the `on` condition with `lv.dataset_id = ds.dataset_id` as the `recalc_dataset_view` is doing.
https://github.com/Hyperfoil/Horreum/blob/a763314753afe7c01890146f74a1f3667c5c221a/horreum-backend/src/main/resources/db/changeLog.xml#L3721-L3736

**To Reproduce**

1. Startup Horreum
2. Create datasets and some views
3. Drop trigger from db ```drop trigger recalc_dataset_view on view_recalc_queue```
4. Force datasets recalculation
5. In the dataset view filter by label value

**Additional Notes**
- Why was it working before? My guess is that it was working by luck.. as the `calc_dataset_view` was inserting wrong data that was later overridden by `recalc_dataset_view`
- Starting from 0.16 we removed the recalculation lock and we might have increased the concurrency level that may not gurantee the previous order (just a guess this)
- The issue was present also before 0.16, we have been just lucky that we did not experience it

PR #2202, opened 2024-12-05
Fix calc_dataset_view join with label values

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2201

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
@@ -636,9 +636,6 @@ public class TestServiceImpl implements TestService {
         RecalculationStatus status = new RecalculationStatus(RunDAO.count("testid = ?1 AND trashed = false", testId));
         // we don't have to care about races with new runs
         RecalculationStatus prev = recalculations.putIfAbsent(testId, status);
-        // ensure the recalculation is removed, with this approach we should guarantee
-        // it gets removed even if transaction-level exception occurs, e.g., timeout
-        Util.registerTxSynchronization(tm, txStatus -> recalculations.remove(testId, status));
         if (prev != null) {
             log.infof("Recalculation for test %d (%s) already in progress", testId, test.name);
             return;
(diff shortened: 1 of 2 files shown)


horreum-backend/src/main/resources/db/changeLog.xml
```diff
@@ -4665,4 +4665,26 @@
...
+        <sql>
+            -- still need db procedure until https://hibernate.atlassian.net/browse/HHH-17314 is fixed in quarkus
+            CREATE OR REPLACE PROCEDURE calc_dataset_view(datasetId bigint) AS $$
+            BEGIN
+            WITH view_agg AS (
+                SELECT
+                    vc.view_id, vc.id as vcid, array_agg(DISTINCT label.id) as label_ids, jsonb_object_agg(label.name, lv.value) as value FROM dataset_schemas ds
+                JOIN label ON label.schema_id = ds.schema_id
+                JOIN viewcomponent vc ON vc.labels ? label.name
+                JOIN label_values lv ON lv.label_id = label.id AND lv.dataset_id = ds.dataset_id
```
lampajr, line 4679: this is the only change: adding `lv.dataset_id = ds.dataset_id` in the join condition

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -636,9 +636,6 @@ public void recalculateDatasets(int testId) {
         RecalculationStatus status = new RecalculationStatus(RunDAO.count("testid = ?1 AND trashed = false", testId));
         // we don't have to care about races with new runs
         RecalculationStatus prev = recalculations.putIfAbsent(testId, status);
-        // ensure the recalculation is removed, with this approach we should guarantee
-        // it gets removed even if transaction-level exception occurs, e.g., timeout
-        Util.registerTxSynchronization(tm, txStatus -> recalculations.remove(testId, status));
```
lampajr, line 641: @johnaohara I also had to remove this because the run transformation is executed async making use of `mediator.executeBlocking(..` therefore this `recalculations.remove` was getting called immediately before returning this method but much before the recalculation was completed which is wrong.. and it does **not** prevent multiple recalculation on the same test to happen concurrently

lampajr: I think we should be fairly safe even without that check as we are wrapping the `mediator.transform` with try-finally that should ensure the finally clause to get called even if transaction-level exceptions are thrown:
```java
                mediator.executeBlocking(() -> {
                    int newDatasets = 0;
                    try {
                        newDatasets = mediator.transform(runId, true);
                    } finally {
                        synchronized (status) {
                            status.finished++;
                            status.datasets += newDatasets;
                            if (status.finished == status.totalRuns) {
                                recalculations.remove(testId, status);
                            }
                        }
                    }
                });
```

lampajr:
I discovered that the `calc_dataset_view` is also creating wrong association between the current dataset and views belonging to other tests just because the viewcomponents were not properly filtered out.

Fixed it by adding `AND vc.view_id IN (SELECT view.id FROM view JOIN dataset ON view.test_id = dataset.testid WHERE dataset.id = datasetId)`.

johnaohara (approved):
LGTM

PR #2209, opened 2024-12-10
Remove use of deprecated table

### LINKED ISSUE #1369: Remove UI dependencies on `@patternfly/react-table/deprecated`
**Describe the bug**

Some of the UI components are deprecated in patternfly 5, we should upgrade the components to the latest version, e.g. `@patternfly/react-table/deprecated`

barreiro:
fixes #1369 

I replaced the `Table` with a `DescriptionList` where appropriate

### DIFF
--- a/horreum-web/src/About.tsx
+++ b/horreum-web/src/About.tsx
@@ -2,6 +2,10 @@ import { useState, useEffect } from "react"
 import {
 	Bullseye,
 	Button,
+    DescriptionList,
+    DescriptionListDescription,
+    DescriptionListGroup,
+    DescriptionListTerm,
 	MenuToggle,
 	Modal,
 	Spinner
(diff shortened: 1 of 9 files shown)


willr3:
The only one I'm concerned with is the `About.tsx` change to `DescriptionList`. Have you looked at the before and after at different resolutions and looks ok?

barreiro:
IMO `DescriptionList` is a more appropriate component to use in some situations (and does look very good). In `About.tsx` here is the before and after:

image

image

willr3 (approved):
lgtm

lampajr (approved):
Looks great to me, especially because it fixes the labelValues/dataset comparison overflow when changing the screen size 😄 

@barreiro could you please rebase and fix the conflict as I merged the "Dark theme switch" first? Then it's ready to go imo

barreiro:
> Looks great to me, especially because it fixes the labelValues/dataset comparison overflow when changing the screen size 😄

yeah, I noticed that issue while testing and the fix was to simply wrap the table in a `InnerScrollContainer` component. 

> @barreiro could you please rebase and fix the conflict as I merged the "Dark theme switch" first? Then it's ready to go imo

done

PR #2210, opened 2024-12-11
change API_KEY examples to avoid ansible issues

willr3:
This is in response to issues user experienced using the examples from the docs in ansible automation. Placing the `$API_KEY` variable reference outside of the single quote parameter definition for `curl -H` in ansible did not properly set the request header. It does work if the variable reference is located inside a double quote (`"`). Changing the documented examples will work in bash and avoid providing examples that conflict with ansible's command line processing.


lampajr (approved):
lgtm

Issue #2174, opened 2024-11-18
New user form requires password


johnaohara:
**Describe the bug**

The "Create New User" form requires a password to be set, but for the OIDC (keycloak) backend, this is not a required field as the password in this use case is not managed via Horreum

![image](https://github.com/user-attachments/assets/a4dfbdbc-04fb-48d9-80b3-874d03ae6c98)

lampajr:
Hi @johnaohara,
I was looking at this as I had to create a new User when using SSO as keycloak backen service and I have a couple of comments:

>but for the OIDC (keycloak) backend, this is not a required field as the password in this use case is not managed via Horreum

That's not actually true because it depends on the `horreum.roles.provider` configuration and keycloak url, i.e.,,
- OIDC enabled, `horreum.roles.provider=keycloak` --> we are sending the provided temporary pwd to the backend keycloak when creating the user there (so here Horreum is actually using the provided pwd)
- OIDC enabled, `horreum.roles.provider=database` --> we are not creating the user on the Keycloak, so the pwd is not used
- OIDC disabled, `horreum.roles.provider=database` (basic auth) --> here the pwd is required, but it is not actually temporary as there is no way to change it from UI


**Proposal**
- If OIDC is enabled, let's mark the pwd as NOT required - then it would be up to the admin to decide whether set it or not in according to the real configuration (based on `horreum.roles.provider`) - then the backend when creating the user (if provider set to `keycloak`) will create the pwd if and only if is is not null/empty.
- If OIDC disabled, let's keep the pwd as required

**Alternative solution**
Expose the `horreum.roles.provider` in the `/api/config/keycloak` and rely on that value as well to mark the pwd as required or completely remove it 

wdyt @johnaohara @barreiro ?

PR #2222, opened 2024-12-19
Make user pwd not required with oidc

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2174

## Changes proposed

- [x] Removed duplicated code in the UI to create new user
- [x] Create the pwd in the Keycloak instance if and only if it is not blank
- [x] Mark the pwd as optional in the form when OIDC is enabled


barreiro (approved):
overall I think it's OK.

I'm a bit put off by the "Temporary" wording for the password in the form, but we can address that as another issue.

lampajr:
> I'm a bit put off by the "Temporary" wording for the password in the form, but we can address that as another issue.

yeah I agree with you but I think it could be acceptable for the time being 👍🏻

PR #2225, opened 2025-01-02
remove usage of deprecated Dropdown and Selector components

barreiro:
I replaced with the `SimpleSelect` (and variants) template in most places. There were other options that could be used, but this one was the easier to work with, looked good and also deals well with long lists of options that  overflow.


barreiro:
the only major change in layout is in `LabelSelect.tsx`  where a complex dropdown was used. I moved the filtering so that I could use a template like in other places. Here is the before horreum-dev corp redhat com_test_13(Horreum Docs) and after localhost_8080_(Horreum Docs)

horreum-web/src/domain/runs/JsonPathSearchToolbar.tsx
```diff
@@ -207,9 +187,9 @@ function execQuery(
     switch (type) {
         case "js":
```
lampajr, line 188: Following the same approach below
```suggestion
        case JS_QUERY_LOCAL:
```

lampajr (approved):
Hi @barreiro,

It looks really great to me! I tried to play a little bit with the Horreum and all looks good, I did not spot anything wrong but I'd like another pair of eyes to double check this.

willr3 (approved):
My memory of the UI code is quite outdated. Thank you for bringing the UI closer to the current era :)

lampajr:
Thanks @barreiro for the PR and @willr3 for the additional check 🙏🏻 

Going to merge this one as I think it comes with some great improvements from the usability standpoint!

Issue #2228, opened 2025-01-07
Move `run_`/`dataset_` schemas updates into Horreum logic


lampajr:
**Feature idea.**

Right now some of the run_schemas and dataset_schemas updates are managed by Postgres triggers. Move their logic into the Horreum codebase such that we don't rely on db triggers anymore for those activities.

Affected entities:
|event_object_table|trigger_name|event_manipulation|action_statement|action_timing|
|--------------------|-----------------------|--------------------|-----------------------------------------------|---------------|
|dataset            |ds_after_insert            |INSERT            |EXECUTE FUNCTION ds_after_dataset_insert_func()  |AFTER|
|run                |rs_after_run_untrash      |UPDATE            |EXECUTE FUNCTION rs_after_run_update()            |AFTER|
|run                |rs_after_run_update        |INSERT            |EXECUTE FUNCTION rs_after_run_update()            |AFTER|
|run                |rs_after_run_update        |UPDATE            |EXECUTE FUNCTION rs_after_run_update()            |AFTER|
|schema            |before_schema_update      |UPDATE            |EXECUTE FUNCTION before_schema_update_func()      |BEFORE|

**Describe the solution you'd like**

Move the logic that updates the `run_schemas` and `dataset_schemas` tables and related into Horreum, i.e., do NOT rely on database triggers.

As part of this effort, ensure the above triggers and functions get removed.

**Additional information**

n/a

PR #2203, opened 2024-12-06
Remove view recalc queue

lampajr:
Followup pull request of https://github.com/Hyperfoil/Horreum/pull/2202

This is not meant to be backported.

Fixes https://github.com/Hyperfoil/Horreum/issues/2220

## Changes proposed

Move the `dataset_view` update logic from db to Horreum java logic.

As part of this refactoring the following entities have been removed:
- Tables
   -  view_recalc_queue and:
    
- Triggers
    - dsv_after_delete
    - dsv_after_update
    - recalc_dataset_view
    - dsv_after_insert
    
- Functions:
    - dsv_after_vc_delete_func
    - dsv_after_vc_update_func
    - recalc_dataset_view
    - dsv_after_lv_insert_func

Moreover the `calc_dataset_view(datasetId bigint, viewId bigint DEFAULT NULL)` signature has changed
such that it can be called in two different scenarios, i.e., by dataset or by dataset-view.

## Check List (Check all the applicable boxes) 

- [x] My code follows the code style of this project.
- [ ] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.

PR #2230, opened 2025-01-08
Removing remaining database triggers

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2228
Fixes https://github.com/Hyperfoil/Horreum/issues/2231

## Changes proposed

Moved the logic to update `run_schemas` and `dataset_schemas` into Horreum logic and as a consequence I removed the following entities:

Triggers:
- ds_after_insert
- rs_after_run_untrash
- rs_after_run_update
- before_schema_update

Functions:
- ds_after_dataset_insert_func
- rs_after_run_update
- before_schema_update_func

I kept the procedure `update_run_schemas` in the db as it is always called by the Horreum directly.

Additionally with commit https://github.com/Hyperfoil/Horreum/pull/2230/commits/714223f199ed0cac77e5fec2c65faa21d9d7e392 I removed the remaining triggers on `Run` table:
- after_run_update_non_data, moved logic into Horreum
- delete_run_validations, we are not even allowing Runs to be deleted so no need to keep that

With commit d18fcc7dfb089c86e8ca3fa2904ff6c9d6ad9975 I also removed all triggers associated to the `label` table:
- lv_before_update
- lv_after_update
- recalc_labels

And related functions. Moreover I also deleted the `label_recalc_queue` table as no longer used.

## Preconditions

- [x] This PR is based on https://github.com/Hyperfoil/Horreum/pull/2203, therefore we have to merge that first and then rebase this one.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
@@ -169,8 +169,8 @@ public class DatasetServiceImpl implements DatasetService {
          LEFT JOIN label_values lv ON dataset.id = lv.dataset_id
          LEFT JOIN label ON label.id = label_id
          """;
-
     //@formatter:on
+
     @Inject
     EntityManager em;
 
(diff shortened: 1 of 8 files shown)


lampajr:
> Looks good, the only concern is if we have good enough coverage already in our testsuite with regards to label and label values updates/delete etc. I had a look at the tests and it looks like we are covering quite a bit.

That's a reasonable concern! 

As part of the latest refactoring I improved the coverage for those logics but for sure it can be improved even further.

>Then if we see any anomalies in testing we are aware of this change and can investigate further.

Totally agree on this, I also tried to document as much as I could the methods such that it should (hopefully) be a bit more easier to debug what's going on

PR #2240, opened 2025-01-15
Rephrase What is Horreum?

diegolovison:
## Changes proposed

Rephrase What is Horreum?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,16 +12,7 @@
 ---
 ## What is Horreum?
 
-Horreum is a service for storing performance data and regression analysis.
-
-Please, visit our project website: 
-
-[https://horreum.hyperfoil.io](https://horreum.hyperfoil.io)
-
-for more information.
-
-Horreum is a [Quarkus](https://quarkus.io/) based application which uses
-[Quinoa](https://quarkiverse.github.io/quarkiverse-docs/quarkus-quinoa/dev/) as its [nodejs](https://nodejs.org/en) engine.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,16 +12,7 @@
...
-
-Please, visit our project website: 
-
-[https://horreum.hyperfoil.io](https://horreum.hyperfoil.io)
-
-for more information.
-
-Horreum is a [Quarkus](https://quarkus.io/) based application which uses
-[Quinoa](https://quarkiverse.github.io/quarkiverse-docs/quarkus-quinoa/dev/) as its [nodejs](https://nodejs.org/en) engine.
+[Horreum](https://horreum.hyperfoil.io) is a web-based platform for storing, managing and visualizing the results of your performance experiment. It provides a user-friendly interface for monitoring the experiments, as well as for analyzing and visualizing the results. It also provides an automated way to detect performance anomaly. Horreum is designed to be scalable and can handle large amount of data, making it a popular choice for tracking performance regressions.
```
willr3, line 15: > and visualization

I do not think we should emphasize Horreum as a visualization tool. It has some charting capabilities but it is not positioned as a visualization tool like Grafana and Jupyter notebooks.

> interface for monitoring the experiments

This section might need clarification. My first thought from this sentence is that Horreum would offer life monitoring of experiments but I don't believe it has that feature. Horreum is a repository of performance test results. It has anomaly detection which could be considered monitoring but that is not what first comes to mind when I read the initial sentence. I would also recommend we not use the word `experiments` unless you are referring to the experiments feature inside horreum. There are a few keywords associated with Horreum (`Test`, `Run`, `Label`, ...) and it would likely help users if we try to keep consistent with those keywords.

diegolovison: Hi @willr3 

Thanks for your review. Would you like to rephrase my change with your changes in mind?

Issue #2242, opened 2025-01-16
Unable to import offline runs


lampajr:
**Describe the bug**

Using external datastores to import runs, e.g., using elasticsearch, if the query on the datastores returns more than 10 runs Horreum will fallback to an offline run processing using the `upload-run` queue.

Unfortunately, uploading those run asynchronously is failing with:

```bash
: org.hibernate.exception.SQLGrammarException: could not execute statement [ERROR: new row violates row-level security policy for table "run"] [insert into run (access,data,description,metadata,owner,start,stop,testid,trashed,id) values (?,?,?,?,?,?,?,?,?,?)]
	at org.hibernate.exception.internal.SQLStateConversionDelegate.convert(SQLStateConversionDelegate.java:91)
	at org.hibernate.exception.internal.StandardSQLExceptionConverter.convert(StandardSQLExceptionConverter.java:58)
	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:108)
	at org.hibernate.engine.jdbc.internal.ResultSetReturnImpl.executeUpdate(ResultSetReturnImpl.java:197)
	at org.hibernate.engine.jdbc.mutation.internal.AbstractMutationExecutor.performNonBatchedMutation(AbstractMutationExecutor.java:134)
	at org.hibernate.engine.jdbc.mutation.internal.MutationExecutorSingleNonBatched.performNonBatchedOperations(MutationExecutorSingleNonBatched.java:55)
	at org.hibernate.engine.jdbc.mutation.internal.AbstractMutationExecutor.execute(AbstractMutationExecutor.java:55)
	at org.hibernate.persister.entity.mutation.InsertCoordinatorStandard.doStaticInserts(InsertCoordinatorStandard.java:194)
	at org.hibernate.persister.entity.mutation.InsertCoordinatorStandard.coordinateInsert(InsertCoordinatorStandard.java:132)
	at org.hibernate.persister.entity.mutation.InsertCoordinatorStandard.insert(InsertCoordinatorStandard.java:104)
	at org.hibernate.action.internal.EntityInsertAction.execute(EntityInsertAction.java:110)
	at org.hibernate.engine.spi.ActionQueue.executeActions(ActionQueue.java:644)
	at org.hibernate.engine.spi.ActionQueue.executeActions(ActionQueue.java:511)
[...]

PR #2243, opened 2025-01-16
Make persistRun transactional

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2242

## Changes proposed

After deeper investigations it turned out the bug was introduced with https://github.com/Hyperfoil/Horreum/pull/2172 that basically removed the `@Transactional` annotation around the `persistRun` method, which is the one invoked by the `upload-run` async emitter.

My guess is that, given that there is not explicit transaction wrapping that method the `roleManager.setRoles(String.join(",", runUpload.roles));` instructions is not visible from the rest of the logic (where the run is persisted) resulting in missing roles at the db level.

Changes:
- [x] Reinstate the `@Transactional` to the `persistRun` method
- [x] Force the addRunFromData to produce `text/plain`

### DIFF
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/RunService.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/RunService.java
@@ -222,6 +222,7 @@ public interface RunService {
             @Parameter(name = "description", description = "Run description", example = "AWS runs"),
 
     })
+    @Produces(MediaType.TEXT_PLAIN)
     @APIResponses(value = {
             @APIResponse(responseCode = "202", description = "The request has been accepted for processing. Returns a list of created run IDs if available, "
                     + "or an empty list if processing is still ongoing. Label values and change detection processing " +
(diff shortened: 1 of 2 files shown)


horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/RunService.java
```diff
@@ -222,6 +222,7 @@ Response add(@QueryParam("test") String testNameOrId,
             @Parameter(name = "description", description = "Run description", example = "AWS runs"),
 
     })
+    @Produces(MediaType.TEXT_PLAIN)
```
lampajr, line 225: I had to explictly state the api produces `text/plain` to fix this issue in the UI:
image

johnaohara: atm it just returns an array of ID's for the accepted Runs to process.
we probably should return a properly formatted JSON object, that contains the URI for each run, e.g https://restfulapi.net/http-status-202-accepted/

For now, returning as plain text is OK

johnaohara (approved):
lgtm

horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/RunService.java
lampajr: yeah I agree, I think this is part of a wider refactoring of the API as we have other endpoints that are not really compliant with the REST standards

PR #2257, opened 2025-02-05
Clear id's importing a new test, check for null

### LINKED ISSUE #2249: Importing a test with missingDataRules fails
**Describe the bug**

The `Test` import from JSON is currently failing when `missingDataRules` are provided.

The logic that imports missing data rules is https://github.com/Hyperfoil/Horreum/blob/c11e9a475718f209582a63531e9376207d29e8ac/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java#L386-L395

Looking at this logic there are 2 different issues:

**When the id IS NOT provided**
In this case, if the rule.id is not set the expectation would be to create a new `Rule` but that's not the case as the logic
fails with:

```
2025-01-22 08:25:40,028 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-73) HTTP Request to /api/test/import failed, error id: 1e35caaf-aea2-48e6-a4ff-1729bbe192bc-3: java.lang.IllegalArgumentException: id to load is required for loading
        ...
	at io.hyperfoil.tools.horreum.entity.alerting.MissingDataRuleDAO.findById(MissingDataRuleDAO.java)
	at io.hyperfoil.tools.horreum.svc.AlertingServiceImpl.importMissingDataRules(AlertingServiceImpl.java:388)
	at io.hyperfoil.tools.horreum.svc.AlertingServiceImpl_ClientProxy.importMissingDataRules(Unknown Source)
```
In this case we should check whether `rule.id != null` and only if not null proceed retrieving the obj and so on.

**When the id IS provided**

When the id is provided, the logic tries to `merge` the entity with the existing one.. but from a semantic point of view this does not make sense. The test import will create a new test, and missing data rules are associated to a specific one.. so it does not make sense at all to merge it with another instance that is related to a different test.

**To Reproduce**

1. Import a Test that contains `missingDataRules`
2. Get a failure importing it

**Version**

What is the version of Horreum ?

If you are using a development branch; what is the commit id ?

```
master / 0.16.x
```

and older versions as well.

**Proposal**

My proposal would be to always create new missing data rules (if provided) either ignoring any id that is being provided or even asserting no id is provided and failing otherwise.

stalep:
Fixes https://github.com/Hyperfoil/Horreum/issues/2249

Clears existing id's if we're importing a new test. Also check for null on the id field that caused the import issue.

### DIFF
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/TestExport.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/data/TestExport.java
@@ -48,19 +48,23 @@ public class TestExport extends Test {
         if (experiments != null && !experiments.isEmpty()) {
             for (ExperimentProfile experiment : experiments) {
                 experiment.testId = id;
+                experiment.id = null;
             }
         }
         if (actions != null && !actions.isEmpty()) {
             for (Action action : actions) {
                 action.testId = id;
+                action.id = null;
             }
         }
         if (subscriptions != null) {
             subscriptions.testId = id;
+            subscriptions.id = null;
         }
         if (missingDataRules != null && !missingDataRules.isEmpty()) {
             for (MissingDataRule rule : missingDataRules) {
                 rule.testId = id;
+                rule.id = null;
             }
         }
     }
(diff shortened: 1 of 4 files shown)


stalep:
No sure why the ubuntu ci fails with that import error...

zakkak:
Can we also get a documentation update reflecting this change and documenting how to duplicate tests using export/import? Thanks

lampajr:
> Can we also get a documentation update reflecting this change and documenting how to duplicate tests using export/import? Thanks

Hi @zakkak, thanks for the suggestion and I fully agree.

I created this https://github.com/Hyperfoil/Horreum/issues/2263 to track the effort.

PR #2258, opened 2025-02-06
make sure we format the response as json when status is 202, fixes #2245

### LINKED ISSUE #2245: RunService.addRunFromData() returns `text/plain`
I had to explicitly state the api produces `text/plain` to fix this issue in the UI:
image
even if the upload was working

_Originally posted by @lampajr in https://github.com/Hyperfoil/Horreum/pull/2243#discussion_r1918145904_

atm it just returns an array of ID's for the accepted Runs to process.
we should return a properly formatted JSON object, that contains the URI for each run, e.g https://restfulapi.net/http-status-202-accepted/

stalep:
Make sure to encode 202 responses as json.

### DIFF
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -670,7 +670,7 @@ paths:
             \ of created run IDs if available, or an empty list if processing is still\
             \ ongoing. Label values and change detection processing is performed asynchronously."
           content:
-            text/plain:
+            application/json:
               schema:
                 type: string
                 example: "101,102,103"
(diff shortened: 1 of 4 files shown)


lampajr:
Hi @stalep, I am not that sure this enough because the returned object (if we use `json`) should be a JSON array and not a comma separated string of ids. 

Another thing is that, if we want to change from text to json, I would try to ensure the endpoint always return JSON regardless of the http code. wdyt?

> NOTE: not sure but the UI might be affected by this change, in the past I had to explicitly set the content type to be tet plain otherwise the UI was failing to parse the response

PR #2268, opened 2025-02-17
docs: Add a note that label id is only required when updating label

jhutar:
## Fixes Issue

Looking into https://horreum.hyperfoil.io/openapi/#tag/Schema/operation/addOrUpdateLabel API doc says `id` ("Unique ID for Label") is required. This should be only required when you are updating label, not adding new one. It confused me.

## Changes proposed

Just a tiny docs change.

### DIFF
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -1551,7 +1551,8 @@ paths:
     post:
       tags:
       - Schema
-      description: Save new or update existing Label for a Schema
+      description: Save new or update existing Label for a Schema (Label id only required
+        when updating existing one)
       operationId: addOrUpdateLabel
       parameters:
       - name: schemaId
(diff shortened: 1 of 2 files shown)


lampajr (changes requested):
Hi @jhutar, thanks a lot for the PR and it looks great to me!

Unfortunately the OpenAPI spec is auto-generated during the build, therefore rather than doing the change there you should do it here https://github.com/Hyperfoil/Horreum/blob/85482d8e6a2fcc20def1534233497d2f97734330/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/services/SchemaService.java#L171 in the `SchemaService` definition. Then, if you run `mvn clean install -DskipTest -DskipITs` you will see the same changes reported on the OpenAPI specs

jhutar:
Thank you Andrea! Fixed that and used the Maven command you mentioned to regenerate that API docs. Hope it is OK now.

lampajr (approved):
>Thank you Andrea! Fixed that and used the Maven command you mentioned to regenerate that API docs. Hope it is OK now.

Thanks a lot to you, lgtm!

Issue #2266, opened 2025-02-14
API call /api/test/372/labelValues?limit=XYZ does not limit number of returned objects


jhutar:
**Describe the bug**

Looking into https://horreum.hyperfoil.io/openapi/#tag/Test/operation/labelValues and looks like `limit` param limits number of labels returned at total across all returned object's values.

Normally limit limits number of returned objects.

Screenshot from Grafana for limit=1:

![Image](https://github.com/user-attachments/assets/d8b53dfe-ccce-4814-818a-4c7e493a96ce)

And limit=2:

![Image](https://github.com/user-attachments/assets/59a12c21-8ffd-4eba-8ca7-95aee82212d9)

And limit=90:

![Image](https://github.com/user-attachments/assets/c40e3955-eca8-404f-900c-0f8a08c06d72)

**To Reproduce**

Attempt to call that API with various limit values and observe number of returned objects

**Version**

Current on Prod

PR #2272, opened 2025-02-19
Fix label values limits

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2266

## Changes proposed

With https://github.com/Hyperfoil/Horreum/pull/1977 we changed how the labelValues implementation was done, the main change was that we moved the run/datasset aggregation from db logic to Java for performance reasons.

Unfortunately this change introduced the regression described in https://github.com/Hyperfoil/Horreum/issues/2266.

With this change I am proposing to move the `limit` logic to Java as well, given that it would be quite not possible to do that at db level given that the aggregation is not there anymore.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/LabelValuesService.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/LabelValuesService.java
@@ -56,7 +56,7 @@ public class LabelValuesService {
                      LEFT JOIN label ON label.id = lv.label_id
                      WHERE dataset.testid = :testId
                         AND (label.id IS NULL OR (:filteringLabels AND label.filtering) OR (:metricLabels AND label.metrics)) INCLUDE_EXCLUDE_PLACEHOLDER
-            ) SELECT * from combined FILTER_PLACEHOLDER ORDER_PLACEHOLDER LIMIT_PLACEHOLDER
+            ) SELECT * from combined FILTER_PLACEHOLDER ORDER_PLACEHOLDER
             """;
 
     protected static final String LABEL_VALUES_QUERY_BY_RUN = """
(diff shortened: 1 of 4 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/LabelValuesService.java
```diff
@@ -531,6 +522,13 @@ protected static List<ExportedLabelValues> parse(List<Object[]> nodes) {
                                     return exportedLabelValue;
                                 })));
 
+        if (limit != null && limit > 0) {
+            page = (page == null || page < 0) ? 0 : page;
+            List<List<ExportedLabelValues>> partitioned = Lists.partition(exportedLabelValues.values().stream().toList(),
```
stalep, line 527: Can we use Collectors.partitioning/groupingBy instead of Guava Lists?

stalep (approved):
Looks good other than that one comment.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/LabelValuesService.java
lampajr: Good suggestion, let me see if that works since it will avoid to add the new guava dependency only for this

lampajr: @stalep looks like I was able to get rid of `guava` as I don't actually need to partition the records.. I just need to skip the first `limit * page` elements and keep the other `limit` ones:
```java
        if (limit != null && limit > 0) {
            long toSkip = (long) limit * ((page == null || page < 0) ? 0 : page);
            return exportedLabelValues.values().stream().skip(toSkip).limit(limit).collect(Collectors.toList());
        }
```

wdyt?

PR #2276, opened 2025-02-20
Upgrade Quarkus and Smallrye openapi plugin

stalep:
Quarkus is updated to the latest version. Had conflicts with bouncycastle 1.80 so only upgraded to 1.79.
json-validator and smallrye-open-api-maven also updated to the latest version.
Did not upgrade keycloak as it introduced some regressions and I do not know how important it is.
KeycloakUserServiceTest.teamTest still fails for me most of the time so not sure if the upgrade caused any regressions for that test. If someone else can verify that would be great.

### DIFF
--- a/horreum-web/src/domain/user/ApiKeys.tsx
+++ b/horreum-web/src/domain/user/ApiKeys.tsx
@@ -109,7 +109,7 @@ export default function ApiKeys() {
     }
 
     const performRenameApiKey = () => {
-        if (renameKeyId) {
+        if (renameKeyId && renameKeyName) {
             userApi.renameApiKey(renameKeyId, renameKeyName)
                 .then(() => void alerting.dispatchInfo("API_KEY_RENAMED", "API key renamed", "API key was successfully renamed", 3000))
                 .catch(error => alerting.dispatchError(error, "API_KEY_NOT_RENAMED", "Failed to rename API key"))
--- a/pom.xml
+++ b/pom.xml
@@ -75,19 +75,22 @@
         <version.maven.install>3.1.4</version.maven.install>
         <version.maven.jar>3.4.2</version.maven.jar>
         <version.maven.javadoc>3.11.2</version.maven.javadoc>
+        <version.maven.helper>3.6.0</version.maven.helper>
         <version.maven.project-info-reports>3.9.0</version.maven.project-info-reports>
         <version.maven.resources>3.3.1</version.maven.resources>
         <version.maven.source>3.3.1</version.maven.source>
+
 
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <commons.math3.version>3.6.1</commons.math3.version>
         <graalvm.version>23.1.2</graalvm.version>
-        <quarkus.version>3.14.4</quarkus.version>
+        <quarkus.version>3.21.0</quarkus.version>
         <quinoa.version>2.5.3</quinoa.version>
         <surefire-plugin.version>3.5.3</surefire-plugin.version>
         <failsafe-plugin.version>3.5.3</failsafe-plugin.version>
         <keycloak-admin-client.version>23.0.3</keycloak-admin-client.version>
+        <quarkus-test-keycloak.version>3.14.4</quarkus-test-keycloak.version>
         <validator.version>1.5.6</validator.version>
         <jayway.jsonpath.version>2.9.0</jayway.jsonpath.version>
 
@@ -217,6 +220,10 @@
                         <groupId>org.keycloak</groupId>
                         <artifactId>keycloak-common</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>org.keycloak</groupId>
+                        <artifactId>keycloak-authz-client</artifactId>
+                    </exclusion>
                 </exclusions>
             </dependency>
 
@@ -235,6 +242,20 @@
                 <groupId>org.keycloak</groupId>
                 <artifactId>keycloak-common</artifactId>
                 <version>${keycloak.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>org.keycloak</groupId>
+                <artifactId>keycloak-authz-client</artifactId>
+                <version>${keycloak.version}</version>
+            </dependency>
+
+            <!-- Force quarkus test keycloak server -->
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-test-keycloak-server</artifactId>
+                <version>${quarkus-test-keycloak.version}</version>
+                <scope>test</scope>
             </dependency>
         </dependencies>
     </dependencyManagement>
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -1,2753 +1,27 @@
 ---
-openapi: 3.0.3
-info:
-  title: Horreum REST API
-  description: "Horreum automated change anomaly detection. For more information,\
-    \ please see [https://horreum.hyperfoil.io/](https://horreum.hyperfoil.io/)"
-  version: "0.18"
-tags:
-- name: Config
-  description: Endpoint providing configuration for the Horreum System
-  x-smallrye-profile-external: ""
-- name: Dataset
-  description: Datasets are used as the basis for all change detection and reporting
-  x-smallrye-profile-external: ""
-- name: Experiment
-  description: Experiments allow users to apply change detection rules to two different
-    datasets. This allows for pass/fail of KPIS based on A/B testing
-  x-smallrye-profile-external: ""
-- name: Run
-  description: Manage test runs. Runs are instances of results of a benchmark execution
-  x-smallrye-profile-external: ""
-- name: Schema
-  description: Manage schemas
-  x-smallrye-profile-external: ""
-- name: Test
-  description: Endpoint giving access to tests defined in Horreum.
-  x-smallrye-profile-external: ""
-- name: User
-  description: Manage user accounts
-  x-smallrye-profile-external: ""
-paths:
-  /api/config/datastore:
-    put:
-      tags:
-      - Config
-      description: Update an existing Datastore definition
-      operationId: updateDatastore
-      requestBody:
-        content:
-          application/json:
-            schema:
-              $ref: "#/components/schemas/Datastore"
-      responses:
-        "200":
-          description: The ID of the updated Datastore
-          content:
-            application/json:
-              schema:
-                format: int32
-                type: integer
-    post:
-      tags:
-      - Config
-      description: Create a new Datastore
-      operationId: newDatastore
-      requestBody:
-        content:
-          application/json:
-            schema:
-              $ref: "#/components/schemas/Datastore"
-      responses:
-        "200":
-          description: The ID for the new Datastore
-          content:
-            application/json:
-              schema:
-                format: int32
-                type: integer
-  /api/config/datastore/types:
-    get:
-      tags:
-      - Config
-      description: Obtain list of available datastore types
-      operationId: datastoreTypes
-      responses:
-        "200":
-          description: OK
-          content:
-            application/json:
-              schema:
-                type: array
-                items:
-                  $ref: "#/components/schemas/TypeConfig"
-  /api/config/datastore/{id}:
-    delete:
-      tags:
-      - Config
-      description: Test a Datastore
-      operationId: deleteDatastore
-      parameters:
-      - name: id
-        in: path
-        required: true
-        schema:
-          format: int32
-          type: integer
-      responses:
-        "204":
-          description: No Content
-  /api/config/datastore/{id}/test:
-    get:
-      tags:
-      - Config
-      description: Test a Datastore connection
-      operationId: testDatastore
-      parameters:
-      - name: id
-        in: path
-        required: true
-        schema:
-          format: int32
-          type: integer
-      responses:
-        "200":
-          description: OK
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/DatastoreTestResponse"
-  /api/config/datastore/{team}:
-    get:
-      tags:
-      - Config
-      description: Obtain list of configured datastores for particular team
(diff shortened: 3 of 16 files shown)


stalep:
There seems to be a regression in the latest version of smallrye-open-api-maven-plugin (or we're doing something wrong with our mapping). I created; https://github.com/smallrye/smallrye-open-api/issues/2207

horreum-web/src/domain/user/ApiKeys.tsx
```diff
@@ -110,7 +110,7 @@ export default function ApiKeys() {
 
     const performRenameApiKey = () => {
         if (renameKeyId) {
-            userApi.renameApiKey(renameKeyId, renameKeyName)
+            userApi.renameApiKey(renameKeyId, renameKeyName!)
```
stalep, line 113: Not sure if this is the "correct" way of doing it, my ts skills are non existent :)

stalep:
I had to change the ordering of some method params since they changed with the smallrye-api upgrade.

lampajr:
> I had to change the ordering of some method params since they changed with the smallrye-api upgrade.

What is worrying me is that how do we know whether there are other changes like these.. for which TS does not complain because the type is still compatible even when switching the params 🤔

lampajr:
> > I had to change the ordering of some method params since they changed with the smallrye-api upgrade.
> 
> What is worrying me is that how do we know whether there are other changes like these.. for which TS does not complain because the type is still compatible even when switching the params 🤔

IIUC the fact that the generated TS files have some params in different order I think that's because the `@RequestBody` https://github.com/microprofile/microprofile-open-api/blob/6ddd1b1ebd85a0b804ad5ae0c06db1a2fe3f879a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/parameters/RequestBody.java#L66 is not required by default differently from version `3.x` where it was NOT required by default.

This difference is quite impacting, at least for all body that we were not marking as required

lampajr:
> This difference is quite impacting, at least for all body that we were not marking as required

Ok I did a check comparing the generated API in master and from this one, and I can confirm the changes you did are the only required one. Therefore we should be safe.

horreum-web/src/domain/user/ApiKeys.tsx
```diff
@@ -110,7 +110,7 @@ export default function ApiKeys() {
 
     const performRenameApiKey = () => {
         if (renameKeyId) {
```
lampajr, line 112: Instead of adding `!` that could fail bad if that is actually undefined, I would rather add the check in the if
```suggestion
        if (renameKeyId && renameKeyName) {
```

horreum-web/src/domain/user/ApiKeys.tsx
lampajr: I would rather add the check in the if stmt rather then adding `!` which could fail bad if the value is actually undefined

lampajr (changes requested):
Thanks @stalep! It looks really great and working so far, I left a couple of suggestions but lmk what you think about them.

>IIUC the fact that the generated TS files have some params in different order I think that's because the @RequestBody https://github.com/microprofile/microprofile-open-api/blob/6ddd1b1ebd85a0b804ad5ae0c06db1a2fe3f879a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/parameters/RequestBody.java#L66 is not required by default differently from version 3.x where it was NOT required by default.

I can confirm that the main breaking change is this one but we should be safe with the changes you did so far - and the fact that all bodies are now `required` should not be a problem but let's keep an eye on it.

I think that as a consequence of this upgrade there a couple of other things that we can/should do:
- `@RequestBody(required = true)` is redundat and can be replaced by `@RequestBody`
- `@schema(example = ...)`, the `example` has been deprecated and can be replaced by `examples`

Moreover, from the logs (during openapi generation) I can see the following errors:
```
[WARNING] There were issues with the specification, but validation has been explicitly disabled.
Errors: 
	-attribute paths.'/api/test'(post).operationId is repeated
	-attribute paths.'/api/run/bySchema'(get).operationId is repeated
	-attribute paths.'/api/schema/{id}/updateAccess'(post).operationId is repeated
	-attribute paths.'/api/test'(put).operationId is repeated
	-attribute paths.'/api/test/{id}/recalculate'(post).operationId is repeated
	-attribute paths.'/api/test/{id}/updateAccess'(post).operationId is repeated
	-attribute paths.'/api/schema'(post).operationId is repeated
	-attribute paths.'/api/test/{id}/labelValues'(get).operationId is repeated
	-attribute paths.'/api/schema/import'(put).operationId is repeated
	-attribute paths.'/api/run/{id}/labelValues'(get).operationId is repeated
	-attribute paths.'/api/test/{id}'(delete).operationId is repeated
	-attribute paths.'/api/test'(get).operationId is repeated
Warnings: 
	-Unused model: DatastoreType
	-Unused model: TableReportConfig
	-Unused model: ChangeDetectionModelType
	-Unused model: TransformationLog
	-Unused model: ActionLog
	-Unused model: LabelInVariable
	-Unused model: LabelInRule
	-Unused model: ActionType
	-Unused model: LabelInFingerprint
	-Unused model: ReportLog
	-Unused model: TableReport
	-Unused model: LabelInReport
	-Unused model: LabelInView
	-Unused model: TableReportData
	-Unused model: ReportComponent

[WARNING] OpenAPI 3.1 support is still in beta. To report an issue related to 3.1 spec, please kindly open an issue in the Github repo: https://github.com/openAPITools/openapi-generator.
[INFO] Generating with dryRun=false
[INFO] Output directory (/home/alampare/repos/Horreum/horreum-api/target/generated/go) does not exist, or is inaccessible. No file (.openapi-generator-ignore) will be evaluated.
[INFO] OpenAPI Generator: go (client)
[INFO] Generator 'go' is considered stable.
[INFO] Environment variable GO_POST_PROCESS_FILE not defined so Go code may not be properly formatted. To define it, try `export GO_POST_PROCESS_FILE="/usr/local/bin/gofmt -w"` (Linux/Mac)
[INFO] NOTE: To enable file post-processing, 'enablePostProcessFile' must be set to `true` (--enable-post-process-file for CLI).
[ERROR] Type {} not yet supported in openapi-normalizer to process OpenAPI 3.1 spec with multiple types.
[ERROR] Please report the issue via https://github.com/OpenAPITools/openapi-generator/issues/new/.
[ERROR] Type {} not yet supported in openapi-normalizer to process OpenAPI 3.1 spec with multiple types.
[ERROR] Please report the issue via https://github.com/OpenAPITools/openapi-generator/issues/new/.
[ERROR] Type {} not yet supported in openapi-normalizer to process OpenAPI 3.1 spec with multiple types.
[ERROR] Please report the issue via https://github.com/OpenAPITools/openapi-generator/issues/new/.
[WARNING] allOf schema `null` containing multiple types (not model) is not supported at the moment.
[WARNING] allOf schema `null` containing multiple types (not model) is not supported at the moment.
```
Most of them were present even before this upgrade!

@stalep I think that all these follow-up changes can be addressed in a different PR as they also relate to the API improvement (I'll be happy to do it)

stalep:
> > > I had to change the ordering of some method params since they changed with the smallrye-api upgrade.
> > 
> > 
> > What is worrying me is that how do we know whether there are other changes like these.. for which TS does not complain because the type is still compatible even when switching the params 🤔
> 
> IIUC the fact that the generated TS files have some params in different order I think that's because the `@RequestBody` https://github.com/microprofile/microprofile-open-api/blob/6ddd1b1ebd85a0b804ad5ae0c06db1a2fe3f879a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/parameters/RequestBody.java#L66 is not required by default differently from version `3.x` where it was NOT required by default.
> 
> This difference is quite impacting, at least for all body that we were not marking as required

Ok, so we should go over these fields and mark them as not required to make sure we're not adding any regressions?
I've not looked at the API where we use all these, but I would assume that fo many where we use RequestBody it should perhaps be required?

lampajr:
>I've not looked at the API where we use all these, but I would assume that fo many where we use RequestBody it should perhaps be required?

IMHO all bodies should be required and looking at the diff between master and this PR there were not many changes, this means that most of the body were already markes as required

horreum-web/src/domain/user/ApiKeys.tsx
lampajr:
```suggestion
            userApi.renameApiKey(renameKeyId, renameKeyName)
```

stalep:
> @stalep I think that all these follow-up changes can be addressed in a different PR as they also relate to the API improvement (I'll be happy to do it)

Awesome! Let's do it that way.

stalep:
Squashed your fixes + updated to smallrye-openapi to 4.0.8, need to have a look at the IT issue.

stalep:
I tested locally updating Keycloak to 26.0.4. That gives a different error when running the IT test:
```
[ERROR] Tests run: 8, Failures: 0, Errors: 1, Skipped: 7, Time elapsed: 24.67 s <<< FAILURE! -- in io.hyperfoil.tools.horreum.it.HorreumClientIT
[ERROR] io.hyperfoil.tools.horreum.it.HorreumClientIT.testApiKeys -- Time elapsed: 0.005 s <<< ERROR!
java.lang.RuntimeException:
io.smallrye.config.ConfigValidationException: Configuration validation failed:
        SRCFG00050: horreum.dev-services.postgres.ssl-certificate in SysPropConfigSource does not map to any root
        SRCFG00050: horreum.dev-services.postgres.ssl-certificate-key in SysPropConfigSource does not map to any root
        at io.quarkus.test.junit.QuarkusIntegrationTestExtension.throwBootFailureException(QuarkusIntegrationTestExtension.java:340)
        at io.quarkus.test.junit.QuarkusIntegrationTestExtension.beforeEach(QuarkusIntegrationTestExtension.java:114)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
Caused by: io.smallrye.config.ConfigValidationException: Configuration validation failed:
        SRCFG00050: horreum.dev-services.postgres.ssl-certificate in SysPropConfigSource does not map to any root
        SRCFG00050: horreum.dev-services.postgres.ssl-certificate-key in SysPropConfigSource does not map to any root
        at io.smallrye.config.SmallRyeConfig.buildMappings(SmallRyeConfig.java:141)
        at io.smallrye.config.SmallRyeConfig.<init>(SmallRyeConfig.java:95)
        at io.smallrye.config.SmallRyeConfigBuilder.build(SmallRyeConfigBuilder.java:773)
        at io.quarkus.deployment.configuration.BuildTimeConfigurationReader.initConfiguration(BuildTimeConfigurationReader.java:427)
        at io.quarkus.deployment.ExtensionLoader.loadStepsFrom(ExtensionLoader.java:141)
        at io.quarkus.deployment.QuarkusAugmentor.run(QuarkusAugmentor.java:107)
        at io.quarkus.runner.bootstrap.AugmentActionImpl.runAugment(AugmentActionImpl.java:368)
        at io.quarkus.runner.bootstrap.AugmentActionImpl.performCustomBuild(AugmentActionImpl.java:156)
        at io.quarkus.test.common.RunCommandLauncher.tryLauncher(RunCommandLauncher.java:53)
        at io.quarkus.test.junit.QuarkusIntegrationTestExtension.doProcessStart(QuarkusIntegrationTestExtension.java:282)
        at io.quarkus.test.junit.QuarkusIntegrationTestExtension.ensureStarted(QuarkusIntegrationTestExtension.java:167)
        at io.quarkus.test.junit.QuarkusIntegrationTestExtension.beforeAll(QuarkusIntegrationTestExtension.java:127)
        ... 1 more

[INFO]
[INFO] Results:
[INFO]
[ERROR] Errors:
[ERROR]   HorreumClientIT.testApiKeys » Runtime io.smallrye.config.ConfigValidationException: Configuration validation failed:
        SRCFG00050: horreum.dev-services.postgres.ssl-certificate in SysPropConfigSource does not map to any root
        SRCFG00050: horreum.dev-services.postgres.ssl-certificate-key in SysPropConfigSource does not map to any root
[INFO]

```

barreiro:
The root cause seems to be quarkusio/quarkus#45432 (introduced in quarkus 3.18) and the fact that we set the keycloak version on `pom.xml` https://github.com/Hyperfoil/Horreum/blob/5ed1ad66c95027282451548e5cd3b46ff2417d27/pom.xml#L97

Update keycloak seems to cause other issues, so I suggest setting the version of `quarkus-test-keycloak-server` to a compatible version
https://github.com/Hyperfoil/Horreum/blob/5ed1ad66c95027282451548e5cd3b46ff2417d27/horreum-backend/pom.xml#L192-L193

With that change the Keycloak tests pass

stalep:
@barreiro which version of quarkus-test-keycloak-server?

barreiro:
I did stick with `3.14.4`, but I reckon any version up to `3.17.8`  would do.

pom.xml
```diff
@@ -83,18 +83,18 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <commons.math3.version>3.6.1</commons.math3.version>
         <graalvm.version>23.1.2</graalvm.version>
-        <quarkus.version>3.19.4</quarkus.version>
+        <quarkus.version>3.21.0</quarkus.version>
         <quinoa.version>2.5.3</quinoa.version>
         <surefire-plugin.version>3.5.2</surefire-plugin.version>
         <failsafe-plugin.version>3.5.2</failsafe-plugin.version>
-        <keycloak-admin-client.version>23.0.3</keycloak-admin-client.version>
+        <keycloak-admin-client.version>26.0.4</keycloak-admin-client.version>
```
barreiro, line 90: I think this property can be dropped. I don't see it used anywhere.

stalep:
I'm not sure why the CI failed on that test. Locally everything is green.

lampajr:
> I'm not sure why the CI failed on that test. Locally everything is green.

The same here, previously it failed on another test.. so maybe there are still some flaky ones

lampajr:
@stalep now it looks good, this should at least confirm we may still have some flaky tests here.

Anyway, I plan to review this again by today so we can merge it before Monday 🤞🏻

lampajr (approved):
LGMT!

Tested both locally and in staging. The SSO seems working as before 🚀

PR #2280, opened 2025-02-24
KeycloakConfig retrieval should use the sslContext

diegolovison:
## Changes proposed
All http requests should use the sslContext configured


lampajr (changes requested):
Hi @diegolovison, thanks for the fix.

Overall looks good but can you please add a test that ensure the behavior would be the expected one, if possible? For instance here https://github.com/Hyperfoil/Horreum/blob/d437b3652b8414ab6b5bab878772a0e110df0279/horreum-integration-tests/src/test/java/io/hyperfoil/tools/horreum/it/HorreumClientIT.java where we have all the tests using the HorreumClient

diegolovison:
Let me think about the test.

PR #2286, opened 2025-02-25
Support huge JSON data

diegolovison:
## Changes proposed

Support huge JSON data

### DIFF
--- a/horreum-backend/src/main/resources/application.properties
+++ b/horreum-backend/src/main/resources/application.properties
@@ -250,3 +250,5 @@ quarkus.elasticsearch.devservices.enabled=false
 
 quarkus.transaction-manager.default-transaction-timeout=1h
 quarkus.arc.fail-on-intercepted-private-method=false
+
+quarkus.http.limits.max-body-size=300M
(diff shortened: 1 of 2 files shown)


lampajr (commented):
Hi @diegolovison, I did some tests and the changes are actually fixing the problem of uploading a _big_ run but TBH I don't think we should set such defaults here.

Both can be added where needed and if and only if are needed:
- Server side: `quarkus.http.limits.max-body-size=300M` can be passed as JVM option when starting the server (i.e., we can do that in out internal lab for instance)
- Client side: `org.jboss.resteasy.client.jaxrs.engines.fileUploadInMemoryThreshold` can be set before creating the `HorreumClient`, so that if users need that can do on their own and have a way to configure it. If we hardcode that users CANNOT override that value

diegolovison:
> Server side: quarkus.http.limits.max-body-size=300M can be passed as JVM option when starting the server (i.e., we can do that in out internal lab for instance)

what is the limit on the database?

lampajr:
> > Server side: quarkus.http.limits.max-body-size=300M can be passed as JVM option when starting the server (i.e., we can do that in out internal lab for instance)
> 
> what is the limit on the database?

Don't remember exactly (need to check), but we should try to keep the json objects as smaller as possible for performance reasons.. therefore I would prefer to keep a lower limit by default

lampajr:
As agreed this is going to be updated to become a new guide section or paragraph showing how users can deal with big JSON data. @diegolovison converting this back to draft but feel free to close this one and open a new one if you consider so.

Issue #2317, opened 2025-03-25
The openapi generator is currently logging some errors


lampajr:
**Describe the bug**

Building the project you can see some errors thrown by the openapi generator, specifically:
```bash
[INFO] --- openapi-generator:7.12.0:generate (typescript) @ horreum-api ---
[ERROR] Undefined array inner type for `null`. Default to String.
[ERROR] Undefined array inner type for `null`. Default to String.
[WARNING] There were issues with the specification, but validation has been explicitly disabled.
Errors: 
	-attribute paths.'/api/test'(post).operationId is repeated
	-attribute components.schemas.Target.items is missing
	-attribute paths.'/api/schema'(get).operationId is repeated
	-attribute paths.'/api/schema'(put).operationId is repeated
	-attribute paths.'/api/run/bySchema'(get).operationId is repeated
	-attribute components.schemas.ViewComponent.items is missing
	-attribute paths.'/api/schema/{id}/updateAccess'(post).operationId is repeated
	-attribute paths.'/api/test'(put).operationId is repeated
	-attribute paths.'/api/test/{id}/recalculate'(post).operationId is repeated
	-attribute paths.'/api/test/{id}'(get).operationId is repeated
	-attribute paths.'/api/test/{id}/updateAccess'(post).operationId is repeated
	-attribute paths.'/api/schema'(post).operationId is repeated
	-attribute paths.'/api/test/{id}/labelValues'(get).operationId is repeated
	-attribute paths.'/api/test/{id}/recalculate'(get).operationId is repeated
	-attribute paths.'/api/subscriptions/{testId}'(get).operationId is repeated
	-attribute paths.'/api/subscriptions/{testid}'(post).operationId is repeated
	-attribute paths.'/api/banner'(get).operationId is repeated
	-attribute paths.'/api/schema/import'(put).operationId is repeated
	-attribute paths.'/api/run/{id}/labelValues'(get).operationId is repeated
	-attribute paths.'/api/schema/{id}'(delete).operationId is repeated
	-attribute paths.'/api/test/{id}'(delete).operationId is repeated
	-attribute paths.'/api/run/test'(post).operationId is repeated
	-attribute paths.'/api/test'(get).operationId is repeated
Warnings: 
	-Unused model: ActionType
	-Unused model: DatastoreType
	-Unused model: LabelInFingerprint
	-Unused model: ChangeDetectionModelType
	-Unused model: ReportLog
	-Unused model: LabelInVariable
[...]

PR #2318, opened 2025-03-25
Update the horreum-api definition

lampajr:
This is a big change but it shouldn't have any impact on external users as the API did NOT change.
The UI will be affected by this, therefore we should check that all functionalities are still working.

## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2317

## Changes proposed

- [x] Re-enable the openapi TS client generation validation
- [x] Remove python/go client generation as not needed here
- [x] Update the horreum-api schemas definition
- [x] Update the horreum-api services definition
- [x] Update the UI accordingly
- [x] Removed the `allOf` from all schemas

> NOTE: the exposed API did NOT change!
> NOTE: what would change is the client generation as the method names will change accordingly to the methods names in the horreum-api services

### DIFF
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -172,7 +172,7 @@ paths:
       tags:
       - Dataset
       description: "Retrieve a paginated list of Datasets, with total count, by Schema"
-      operationId: listBySchema
+      operationId: listDatasetsBySchema
       parameters:
       - name: uri
         in: query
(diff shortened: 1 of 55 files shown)


lampajr:
Tested locally and it looks good to me but I'd like to have another pair of eyes checking it 👀

lampajr:
> Removed the `allOf` from all schemas

I did this because given that our source of truth is `horreum-api` module (i.e., java code) I think they are not actually needed, especially because the client generation would be the same even without that `allOf`. With this the generated openapi spec is simpler w/o such allOf

stalep (approved):

PR #2340, opened 2025-04-07
Inherit quarkus-test-keycloak-server from quarkus

lampajr:
Looks like we don't need to specify an older version of `quarkus-test-keycloak-server` to make ITs working.

## Changes proposed

- [ ] Remove the explicit `quarkus-test-keycloak-server` dependency
- [ ] Remove unused `keycloak-admin-client.version` version property


lampajr:
@stalep @barreiro, it looks like that with the current setup we don't need to stick with an older `quarkus-test-keycloak-server` version for our testing. Can you please double check?

stalep:
Looks good, works fine locally.

Issue #845, opened 2023-10-30
Horreum observability


johnaohara:
The Opentelemetry extension allows us to view telemetry data and export it to prometheus

I was thinking about: https://github.com/Hyperfoil/Horreum/issues/342 and observability in general.  I was able to identify the root cause of the issue using custom opentelemetry spans to track the tasks placed in the taskqueue.

How do we want to observe a running prod instance of horreum, want info do we want to track and how do we want to access it?  For example, there is not enough information to understand what is happening in #342 from the logs even at a debug level. 

Do we want to rely on grepping debug logs to find information, or do we want some form of observability tool to be able to query recorded events and obtain the necessary information to understand what is happening the running instance?

It is relatively simple to add OpenTelemetry as a telemetry backend, but the neccesary infrastructure is required to process the telemetry data (otel collector, promethues, jaeger etc).  the flip side is the Quarkus OpenTelemetry plugin provides insight and error tracing in some of the subsystems in Quarkus. Including parts of the system we would not naturally think to instrument, e.g.;

![Screenshot from 2023-02-07 19-00-12](https://user-images.githubusercontent.com/959822/217340419-0dd28713-a6fa-493b-a05b-dcaf55c84ff5.png)

_Originally posted by @johnaohara in https://github.com/Hyperfoil/Horreum/discussions/365_

stalep:
I think this would help us a lot when we encounter issues in Horreum and we have trouble finding the errors in the logs. Even though log reporting should be better with 0.10 this would make it even better.

PR #2341, opened 2025-04-07
Enable default micrometer metrics

lampajr:
## Changes proposed

I think it would be nice to enable the default micrometer metrics to gather Horreum statistics at runtime.

### DIFF
--- a/horreum-backend/pom.xml
+++ b/horreum-backend/pom.xml
@@ -180,6 +180,11 @@
         </dependency>
 
         <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-micrometer-registry-prometheus</artifactId>
+        </dependency>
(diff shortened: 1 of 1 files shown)


stalep:
I think it sounds useful. Should we add some documentation to this such that users know how to connect to it via Prometheus as well?

lampajr:
> @lampajr this relates to : #845
> 
> IDK where we are with the micrometer vs opentelemtry debate now in quarkus

Oh I missed that issue 🤦🏻 we can discuss further about what approach we should follow, TBH I don't have a strong opinion on this topic.

> I think it sounds useful. Should we add some documentation to this such that users know how to connect to it via Prometheus as well?

Yeah maybe we could add some very simple documentation about the exposed metrics and how users could get get them.

Issue #2343, opened 2025-04-09
Cannot create new actions under Test configuration


lampajr:
**Describe the bug**

Currently it is not possible to create new Actions under Test configuration

![Image](https://github.com/user-attachments/assets/1d1636fa-7732-48a4-b6b4-b068aa28603e)

Full stacktrace:

```bash
jakarta.persistence.OptimisticLockException: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect): [io.hyperfoil.tools.horreum.entity.data.ActionDAO#-1]
	at org.hibernate.internal.ExceptionConverterImpl.wrapStaleStateException(ExceptionConverterImpl.java:209)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:95)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:167)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:173)
	at org.hibernate.internal.SessionImpl.fireMerge(SessionImpl.java:865)
	at org.hibernate.internal.SessionImpl.merge(SessionImpl.java:840)
	at io.quarkus.hibernate.orm.runtime.session.TransactionScopedSession.merge(TransactionScopedSession.java:169)
	at org.hibernate.engine.spi.SessionLazyDelegator.merge(SessionLazyDelegator.java:272)
	at org.hibernate.Session_OpdLahisOZ9nWRPXMsEFQmQU03A_Synthetic_ClientProxy.merge(Unknown Source)
	at io.hyperfoil.tools.horreum.svc.ActionServiceImpl.merge(ActionServiceImpl.java:201)
	at io.hyperfoil.tools.horreum.svc.ActionServiceImpl.update(ActionServiceImpl.java:232)
	at io.hyperfoil.tools.horreum.svc.ActionServiceImpl_Subclass.update$$superforward(Unknown Source)
	at io.hyperfoil.tools.horreum.svc.ActionServiceImpl_Subclass$$function$$3.apply(Unknown Source)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:73)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext$NextAroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:97)
[...]

PR #2344, opened 2025-04-09
Separate global and test action endpoints

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2343

## Changes proposed

- [x] New endpoint for adding global actions `api/action/global`
- [x] Global actions require ADMIN role
- [x] Test actions require TESTER role
- [x] Add or update test Actions accordingly


lampajr:
There are still a couple of test errors, need to check them

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
```diff
@@ -160,27 +161,52 @@ void validate(Action action) {
     @WithRoles
     @Transactional
     @Override
-    public Action add(Action action) {
+    public Action addGlobalAction(Action action) {
```
johnaohara, line 164: Actually, do we want to restrict this to Admin role?

lampajr: It is already, that restriction is truncated in the diff changes 😄 but if you extend it on top you would see it
image

Issue #2347, opened 2025-04-11
Webhooks are not forwarding the events


lampajr:
**Describe the bug**

It has been reported that even if the webhooks configuration is correct, the events are not actually posted to the configured target system.

**To Reproduce**

1. Start Horreum locally
2. Start a dummy server, e.g., `podman run -d --rm -P mockserver/mockserver`
3. Create a new allowed site to `http://localhost:$PORT` where `$PORT` is the mockserver port
4. Create a new Test and add a new Action for `run/new` as generic HTTP post
5. Try to upload a run and you'll see that your server is not getting called 

**Version**

What is the version of Horreum ?

If you are using a development branch; what is the commit id ?

```
0.17.1, still present in main
```

lampajr:
After deep investigation I found out the root cause:

Basically the main problem is that there is an issue in how the actions are retrieved for a specific test, see https://github.com/Hyperfoil/Horreum/blob/b4de5a719ba455c26034f3b7582a591f20eceddc/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java#L70 https://github.com/Hyperfoil/Horreum/blob/b4de5a719ba455c26034f3b7582a591f20eceddc/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java#L298-L304

* It is calling the `getActions` method providing as event the `AsyncEventChannels.name()` which is converted to something like `RUN_NEW` (in the case of new run event) which is literally the string version of the enum
* Unfortunately, events are saved in the db using a different conversion (e.g., `RUN_NEW` --> `run/new`) 
* Thus, the list of actions will always be empty as no actions will match `RUN_NEW`!

PR #2348, opened 2025-04-11
Create custom action events enum

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2347

## Changes proposed

- [x] Create ad-hoc `ActionEvent` enum with proper db mapping
- [x] Don't use `AsyncEventChannel` as webhook events

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionEvent.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionEvent.java
@@ -1,0 +1,33 @@
+package io.hyperfoil.tools.horreum.svc;
+
+public enum ActionEvent {
+    TEST_NEW("test/new"),
+    RUN_NEW("run/new"),
+    CHANGE_NEW("change/new"),
+    EXPERIMENT_RESULT_NEW("experiment_result/new"),;
+
+    // this represents the value stored in the database
+    private final String value;
+
+    ActionEvent(String value) {
+        this.value = value;
+    }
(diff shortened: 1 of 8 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionEvent.java
johnaohara, line 1: Can we not use the existing enum, and if needed add a boolean flag to indicate if the event is a webhook event? We will have 2 classes to maintain if the events change

lampajr: We could use the same but IMHO they have different meanings and usages, therefore I think that having separate enums is easier to maintain. Moreover the AsyncEventChannel is mostly used for testing purposes.. so I think we could probably get rid of it at some point. wdyt?

johnaohara: OK, we should look at removing AsyncEventChannel, and updating the test infra to work without that Class

johnaohara (approved):

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionEvent.java
lampajr: Created https://github.com/Hyperfoil/Horreum/issues/2349 to track the issue

Issue #2362, opened 2025-04-29
Quarkus builder exception with "wrong usage(s) of @Blocking"


lampajr:
**Describe the bug**

I am trying to create a Quarkus application that uses the `horreum-client` dependency but the mvn build is failing with:
```
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:3.12.3:build (default) on project horreum-github-app: Failed to build quarkus application: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
[ERROR] 	[error]: Build step io.quarkus.deployment.execannotations.ExecutionModelAnnotationsProcessor#check threw an exception: java.lang.IllegalStateException: 
[ERROR] Wrong usage(s) of @Blocking found:
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.getRoles()
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.searchUsers(java.lang.String)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.info(java.util.List<java.lang.String>)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.createUser(io.hyperfoil.tools.horreum.api.services.UserService$NewUser)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.removeUser(java.lang.String)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.getTeams()
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.defaultTeam()
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.setDefaultTeam(java.lang.String)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.teamMembers(java.lang.String)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.updateTeamMembers(java.lang.String, java.util.Map<java.lang.String, java.util.List<java.lang.String>>)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.getAllTeams()
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.addTeam(java.lang.String)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.deleteTeam(java.lang.String)
[ERROR] 	- io.hyperfoil.tools.horreum.api.services.UserService.administrators()
[...]

PR #2363, opened 2025-04-29
Removed @Blocking from UserService

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2362

## Changes proposed

- [ ] Remove the @Blocking annotation from `UserService`


johnaohara (approved):
I think it is fine to remove the `@Blocking` annotation here, the endpoint will default to dispatch to the worker pool based on the method signature: https://quarkus.io/blog/resteasy-reactive-smart-dispatch/#new-world-new-rules

If we want to retain the explicit definition, the impl class methods should be annotated with `@Blocking`, but this behaviour will not change unless Quarkus changes the default behaviour

PR #2364, opened 2025-04-29
Update the HorreumClient Run service

lampajr:
This aims to include all methods from originl RunService but the addRunFromData that has FileUpload as param

I noticed that the HorreumClient is overriding the `RunService` because of the `addRunFromData`, which is leaking some reactive objects (i.e., `FileUpload`). Thus, the RunService defined in the client is not up to date wrt to the latest changes and improvements.

I am proposing to update so that we include all methods in the client RunService (e.g., get Run label values was missing).

## Changes proposed

- [x] Re-define the `client.RunService` to include all methods from `services.RunService` but `addRunFromData` which is overridden by a new version that takes data/metadata as `JsonNode` type
- [x] Adapt the method names to the new ones, as defined in the `services.RunService`

### DIFF
--- a/horreum-client/src/main/java/io/hyperfoil/tools/RunServiceExtension.java
+++ b/horreum-client/src/main/java/io/hyperfoil/tools/RunServiceExtension.java
@@ -18,6 +18,7 @@ import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import io.hyperfoil.tools.horreum.api.SortDirection;
 import io.hyperfoil.tools.horreum.api.client.RunService;
 import io.hyperfoil.tools.horreum.api.data.Access;
+import io.hyperfoil.tools.horreum.api.data.ExportedLabelValues;
 import io.hyperfoil.tools.horreum.api.data.Run;
 import io.hyperfoil.tools.horreum.api.services.RunService.RunCount;
 import io.hyperfoil.tools.horreum.api.services.RunService.RunSummary;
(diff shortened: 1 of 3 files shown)


johnaohara (approved):
Do we know why the api changes did not cause errors in the test? 

there are calls to incorrect `delegate.` object methods, did they not throw compilation errors?

lampajr:
> Do we know why the api changes did not cause errors in the test?
> 
> there are calls to incorrect `delegate.` object methods, did they not throw compilation errors?

The problem is that the `delegate` was (and is) a completely different `RunService`.. therefore from a compilation perspective everything was good.. and as long as the underlying REST call was working no issue would have appeared.

I think that in the long-term we should try to remove such override/replacement of `RunService` in the Client and use directly the horreum-api RunService as we are doing for all the other services

johnaohara:
> I think that in the long-term we should try to remove such override/replacement of RunService in the Client and use directly the horreum-api RunService as we are doing for all the other services 

I agree, but at the moment, `org.jboss.resteasy.reactive.multipart.FileUpload` pulls in a dependency on `resteasy-reactive`, which pulls in all sorts of dependencies that we don't want. The service relies on `FileUpload` for validation, so we need to refactor the service to expose any classes from `resteasy-reactive`

lampajr:
@johnaohara yeah exactly it would require a refactoring on the api side but I think this would be still the "best" solution in the long-term (will remove the maintenance of this addition RunService replacement).

PR #2451, opened 2025-08-07
draft

diegolovison:


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
```diff
@@ -127,6 +127,12 @@ public void onNewTest(Test test) {
         executeActions(ActionEvent.TEST_NEW, test.id, test, true);
     }
 
+    @WithRoles(extras = Roles.HORREUM_SYSTEM)
+    @Transactional
+    public void onDatasetLabelsComputed(Integer testId, Object payload) {
```
lampajr, line 132: I am expecting this to be of type `Dataset.EventNew`
```suggestion
    public void onDatasetLabelsComputed(Integer testId, Dataset.EventNew payload) {
```

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
```diff
@@ -127,6 +127,12 @@ public void onNewTest(Test test) {
...
+        executeActions(ActionEvent.DATASET_LABELS_COMPUTED, testId, payload, true);
```
lampajr, line 133: Here, I am expecting the `payload` to be the `Dataset` obj. So from the `Dataset.EventNew` retrieve the actual dataset object and pass it as payload

horreum-web/src/domain/actions/reducers.ts
```diff
@@ -85,6 +86,23 @@ export const testEventTypes = [
...
+    "owner": "dev-team",
+    "access": 0,
+    "schema": {
+      "1": "urn:my-schema:1.0"
+    },
+    "testid": 10,
+    "testname": "My test",
+    "description": null,
+  }`,
+    ],
```
lampajr, line 105: The object should match the `Dataset` object structure, this is actually just an example with fake data

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ServiceMediator.java
```diff
@@ -172,6 +172,7 @@ void onNewDataset(Dataset.EventNew eventNew) {
     public void processDatasetEvents(Dataset.EventNew newEvent) {
         onNewDataset(newEvent);
         validateDataset(newEvent.datasetId);
+        this.actionService.onDatasetLabelsComputed(newEvent.testId, newEvent);
```
lampajr, line 175: I am still not 100% sure to have it here, but at the moment let's keep it

lampajr (changes requested):
Thanks @diegolovison , I left a couple of suggestions but overall the changes look good.

PR #2455, opened 2025-08-08
Add "Export to CSV" button in DatasetComparison

### LINKED ISSUE #2433: Export .csv from dataset compare
**Feature idea.**

There is the possibility to compare datasets and export that comparison as .pdf. In cases where the goal is to further process that data, in ways that are not done easily in Horreum, it would be nice to export as .csv file. That way, it would be easy to import that data into other tools.

**Describe the solution you'd like**

Add a button "Export to CSV" right next to the "Export to PDF" that would save the comparison as a .csv file.

Image

barreiro:
Fixes #2433

### DIFF
--- a/horreum-web/src/domain/runs/DatasetComparison.tsx
+++ b/horreum-web/src/domain/runs/DatasetComparison.tsx
@@ -1,14 +1,16 @@
 import { SetStateAction, useContext, useEffect, useMemo, useRef, useState } from "react"
 import { useSelector } from "react-redux"
 import {
-    ActionGroup,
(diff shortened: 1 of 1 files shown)


lampajr:
Thanks @barreiro , looking into this 👀

barreiro:
fixed the `sanitizer` function to handle `\n` character in labels and values

lampajr:
@barreiro @diegolovison can we proceed merging this one?

diegolovison:
Yes

barreiro:
I think so

PR #2459, opened 2025-08-12
Add new Dataset Labels Computed event

diegolovison:
## Fixes Issue

https://issues.redhat.com/browse/MPL-960

**Problem**
At the moment users can get notified when a new Run is uploaded, but given the async nature of the run upload logic (which means that label values are calculated asynchronously after some time) they cannot rely on such event to retrieve label values data.

**Proposal**
Introduce a new event "new/dataset" that, if subscribed (through actions config), users can get notifed when a "dataset" is completely computed - this means that this should be triggered when label values for a specific datasets are finalized (or recalculated?)

**Acceptance criteria**
Introduce additional "new/dataset" event
Ensure configured actions gets called when all label values are computed for that dataset

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
@@ -23,6 +23,7 @@ import io.hyperfoil.tools.horreum.api.SortDirection;
 import io.hyperfoil.tools.horreum.api.alerting.Change;
 import io.hyperfoil.tools.horreum.api.data.Action;
 import io.hyperfoil.tools.horreum.api.data.AllowedSite;
+import io.hyperfoil.tools.horreum.api.data.Dataset;
 import io.hyperfoil.tools.horreum.api.data.Run;
 import io.hyperfoil.tools.horreum.api.data.Test;
 import io.hyperfoil.tools.horreum.api.data.TestExport;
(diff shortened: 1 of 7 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
```diff
@@ -127,6 +130,21 @@ public void onNewTest(Test test) {
         executeActions(ActionEvent.TEST_NEW, test.id, test, true);
     }
 
+    @WithRoles(extras = Roles.HORREUM_SYSTEM)
```
lampajr, line 133:
```suggestion
    @WithRoles(extras = Roles.HORREUM_SYSTEM)
    @Transactional
```

lampajr (changes requested):
Thanks a lot @diegolovison, especially for the test!!

 The PR looks great, there are just a couple of things to fix (minors):
- Adding `@Transactional` to be safer
- Commit the uncommited changes (as you can see the CI is failing, it should be enough to run `mvn clean package -DskipTests -DskipITs` and commit the automated changes)

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ActionServiceImpl.java
diegolovison: - Added `@Transactional`

- No idea by what you mean with "Commit the uncommited changes"

diegolovison:
@lampajr it is green

PR #2468, opened 2025-08-26
Patternfly 6.3

barreiro:
## Changes proposed

This PR updates Patternfly to 6.3 version.

There are a few style changes to make the UI more 'clean'

### DIFF
--- a/horreum-web/src/domain/actions/ActionList.tsx
+++ b/horreum-web/src/domain/actions/ActionList.tsx
@@ -1,7 +1,7 @@
 import {useContext, useEffect, useMemo, useState} from "react"
 import { useSelector } from "react-redux"
 
-import { Button, Hint, HintBody, Switch, Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core"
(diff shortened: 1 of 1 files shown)


lampajr (changes requested):
Hi @barreiro, overall looks great!!!

I noticed a couple of things that IMHO could be improved a little bit, they are all minors so nothing actually breaking but I believe it would still be nice to include:

1. Dataset comparison table does not have borders and this could be a bit confusing to users
image
I would like to clearly separate the comparison table from the datasets list but open to discuss

2. Increase the schemas list padding/margin in the dataset view, especially when there are multiple schemas associated (they are too close to the buttons underneath)
image

3. In the schema label definition, the helper buttons for `filtering`/`metrics` option should just contain the question mark icon w/o the button borders
image
The same is showed in change detection tab, therefore it could be a widespread issue
image

4. In the report I would change the background color for the comments, I would stick with the color scheme we are using, e.g., using the gray instead of pure black (but that's just my opinion here, open to other alternatives)
image

5. In the administration tab / global actions increase the padding/margin for the comment/banner
image

barreiro:
@lampajr thanks for reviewing. I  addressed your 5 comments.

Issue #2401, opened 2025-06-02
When 2 labela are created/modified (?), all datasets in the test are recalculated 2 times


jhutar:
**Describe the bug**

Recently I created ~30 labels using API script (so one by one - this is in contrast to how you do it in UI where you can create many of them and save only once). his schema was used in a test with 9500 results. This generated 285000 dataset recalculation events in the queue that made calculating label values for new uploads (to any test) significantly delayed.

This was discussed in https://redhat-internal.slack.com/archives/C069K4Q15JB/p1748284489461919

Maybe when new event is added to recalculation queue, it should only be added when there is no such event already?

Andrea also mentioned Horreum team is looking into adding a separate queue only for new results.

**To Reproduce**

Create 2 labels via the API and notice all datasets of a given test gets recalculated two times (I assume this can only be checked in Horreum logs).

**Version**

Current version in prod.

PR #2472, opened 2025-09-02
updated quarkus to 3.26.3

stalep:
This fails atm with;
```
[ERROR] Caused by: java.lang.NoClassDefFoundError: io/quarkus/vertx/http/runtime/HttpBuildTimeConfig
[ERROR]         at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
[ERROR]         at java.base/java.lang.Class.privateGetDeclaredMethods(Class.java:3578)
[ERROR]         at java.base/java.lang.Class.getDeclaredMethods(Class.java:2676)
[ERROR]         at io.smallrye.config.ConfigMappingInterface.createConfigurationInterface(ConfigMappingInterface.java:783)
[ERROR]         at io.smallrye.config.ConfigMappingInterface$1.computeValue(ConfigMappingInterface.java:42)
[ERROR]         at io.smallrye.config.ConfigMappingInterface$1.computeValue(ConfigMappingInterface.java:40)
[ERROR]         at java.base/java.lang.ClassValue.getFromHashMap(ClassValue.java:229)
[ERROR]         at java.base/java.lang.ClassValue.getFromBackup(ClassValue.java:211)
[ERROR]         at java.base/java.lang.ClassValue.get(ClassValue.java:117)
[ERROR]         at io.smallrye.config.ConfigMappingInterface.getConfigurationInterface(ConfigMappingInterface.java:93)
[ERROR]         at io.smallrye.config.ConfigMappingLoader.loadImplementation(ConfigMappingLoader.java:128)
[ERROR]         at io.smallrye.config.ConfigMappingLoader$1.computeValue(ConfigMappingLoader.java:28)
[ERROR]         at io.smallrye.config.ConfigMappingLoader$1.computeValue(ConfigMappingLoader.java:25)
[ERROR]         at java.base/java.lang.ClassValue.getFromHashMap(ClassValue.java:229)
[ERROR]         at java.base/java.lang.ClassValue.getFromBackup(ClassValue.java:211)
[ERROR]         at java.base/java.lang.ClassValue.get(ClassValue.java:117)
[ERROR]         at io.smallrye.config.ConfigMappingLoader.configMappingProperties(ConfigMappingLoader.java:67)
[ERROR]         ... 63 more
[ERROR] Caused by: java.lang.ClassNotFoundException: io.quarkus.vertx.http.runtime.HttpBuildTimeConfig
[...]

PR #2483, opened 2025-09-16
Create and update schema labels in bulk

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2401

## Changes proposed

When multiple labels are updated/created/deleted, issue one single dataset event that will re-compute only those label values related to the labels that have been updated, and after that triggers the change detection recalculation only **once**


lampajr:
Converting to draft as there are other changes I need to implement as part of this PR, e.g., batching the label values calculation

lampajr:
> Regarding the API change to List vs single object makes sense with the internal changes, no need to keep it backwards compatible imo since we're updating this in a new version. Also +1 to separate update/create/delete endpoints!

Sounds good then!

I will implement the "label values calculation in batch" and then I will mark this PR as ready again

lampajr:
> I will implement the "label values calculation in batch" and then I will mark this PR as ready again

Implemented ✅ 

With this new impl it will batch multiple label values computation in the same query, this should reduce the amount of queries we perform per event (and only one change detection recalc per dataset event)

lampajr:
@stalep I am reviewing https://github.com/Hyperfoil/Horreum/pull/2472 and if everything is good I will merge that one first and rebase this PR to check if everything is still fine

PR #2487, opened 2025-09-17
Fix change detection tooltip

### LINKED ISSUE #2486: Text not displayed on change popup in dark mode
**Describe the bug**
When using dark mode the text in the Change popup is not shown:

Image

The same is working as expected when using light mode

Image

**To Reproduce**

Open a test with some changes and go over those using light mode and dark mode and you will see what reported above.

**Version**

What is the version of Horreum ?

```master``` branch after merge of https://github.com/Hyperfoil/Horreum/pull/2468

barreiro:
Fixes #2486


lampajr (approved):
lgtm

PR #2488, opened 2025-09-17
React 19

barreiro:
this pr updates react to version 19 and the necessary dependencies. 

`patternfly-templates` does not support the update, so I'm bringing those components that Horreum uses into the codebase, for now.

`react-table` update has breaking changes on the API, so there are major changes in `CustomTable`and it's usages.


horreum-web/src/store.ts
lampajr, line 1: Are we loosing something with this change?

barreiro: I don't think so. I followed the guide https://redux.js.org/usage/migrating-to-modern-redux#store-setup-with-configurestore

PR #2494, opened 2025-09-18
updates to quarkus, maven plugins and bouncycastle:bcpkix

stalep:
org.bouncycastle:bcprov-jdk18on 1.82 seems to still depend on bcpkix 1.81 so leaving it as it is.


lampajr (approved):
lgtm

PR #2514, opened 2025-09-25
[0.18] Create and update schema labels in bulk

rh-appservices-perf:
**Backport:** https://github.com/Hyperfoil/Horreum/pull/2483

## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2401

## Changes proposed

When multiple labels are updated/created/deleted, issue one single dataset event that will re-compute only those label values related to the labels that have been updated, and after that triggers the change detection recalculation only **once**

**Breaking change**
- Create/update label endpoints now expect a list of labels rather than only one label

### DIFF
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -4613,9 +4613,8 @@ paths:
           description: Created
   /api/schema/{schemaId}/labels:
     put:
-      description: Update existing Label for a Schema (Label id only required when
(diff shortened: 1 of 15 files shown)


lampajr:
I am trying to backport this improvement in the current stable branch, so that if needed we can roll-out a bug fixing release for `0.18`

lampajr:
Nevermind, it is changing the API which is a breaking change. Closing..

PR #2516, opened 2025-09-26
Improve run change detection process

lampajr:
This PR aims to improve in some part the change detection re-execution during dataset recalculation.

## Changes proposed

- [x] Re-introduce the concept of `fingerprint hash` 
- [x] Use the hash to perform comparison based on fingerprint instead of jsonb comparisons

> NB: Even if at the moment in prod we don't have hashes collision, this code does not handle that case

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/entity/FingerprintDAO.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/entity/FingerprintDAO.java
@@ -7,6 +7,8 @@ import jakarta.persistence.Id;
 import jakarta.persistence.JoinColumn;
 import jakarta.persistence.MapsId;
 import jakarta.persistence.OneToOne;
+import jakarta.persistence.PrePersist;
+import jakarta.persistence.PreUpdate;
 
 import org.hibernate.annotations.Type;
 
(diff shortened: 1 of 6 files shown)


lampajr:
I did some _local_ tests, and with the improvements we should be able to reduce the allocation size and, as a consequence, the GC pressure, see some tests below.

**Current prod**

Screenshot From 2025-09-26 16-54-16

**Hash and jsob comparison**

Screenshot From 2025-09-26 16-54-21

**Only hash comparison**

Screenshot From 2025-09-26 17-39-41

The test was the same on all three executions: updating 2 labels on `specj-web-hyperfoil-v2` schema one at the time, i.e., clicking save 2 times.

stalep:
> I did some _local_ tests, and with the improvements we should be able to reduce the allocation size and, as a consequence, the GC pressure, see some tests below.
> 
> **Current prod**
> Screenshot From 2025-09-26 16-54-16
> 
> **Hash and jsob comparison**
> Screenshot From 2025-09-26 16-54-21
> 
> **Only hash comparison**
> Screenshot From 2025-09-26 17-39-41
> 
> The test was the same on all three executions: updating 2 labels on `specj-web-hyperfoil-v2` schema one at the time, i.e., clicking save 2 times.

That's a huge improvement!

stalep:
Do you mind squashing then we'll merge :)

lampajr:
> Do you mind squashing then we'll merge :)

@stalep squashed right now! ✅

lampajr:
Merging this one to check performances and correctness in out staging env

PR #2517, opened 2025-09-29
Reduce AMQP reconnect interval

lampajr:
Based on the documentation at https://quarkus.io/guides/amqp-reference
```
The interval in second between two reconnection attempts

Type: int
```

We are using a too high value for `amqp-reconnect-interval`, as we are using `10000` that in seconds means `>15 mins`

## Changes proposed

- [x] Reducing the reconnect interval to the default value: `10s`


lampajr:
I am still checking whether that option should be set at `channel` level or global

PR #2522, opened 2025-09-30
Generic datasets processing improvements

lampajr:

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ConfigServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ConfigServiceImpl.java
@@ -35,7 +35,7 @@ public class ConfigServiceImpl implements ConfigService {
     @ConfigProperty(name = "horreum.privacy")
     Optional<String> privacyStatement;
 
-    @ConfigProperty(name = "horreum.application.commit")
+    @ConfigProperty(name = "horreum.application.commit", defaultValue = "<unknown>")
     String appCommit;
 
     @Inject
--- a/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/alerting/Change.java
+++ b/horreum-api/src/main/java/io/hyperfoil/tools/horreum/api/alerting/Change.java
@@ -15,11 +15,14 @@ public class Change {
     @NotNull
     @JsonProperty(required = true)
     public Variable variable;
+
     @JsonIgnore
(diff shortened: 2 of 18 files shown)


lampajr:
@stalep since you were able to reproduce the problem we were seeing, do you mind to check with this branch? I tried to simplify some parts by reducing the expected allocation when really not needed.

E.g., the `Change` does not have the full `Dataset` anymore as it should not be used anywhere.

lampajr:
For tests with plenty of label values, and with big json values in them, most of the CPU was spent in creating the fingerprint (method `createFingerprint`)

image

Looking at the code it seems the logic is not optimized at all, as it retrieve all label values and perform the filtering by label name in Java.. this has the drawback of allocating a lot more garbage that could be useless if the label will be filtered out.

Planning to test this in our staging to see the performances in a more stable environment

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/ConfigServiceImpl.java
```diff
@@ -35,8 +39,8 @@ public class ConfigServiceImpl implements ConfigService {
     @ConfigProperty(name = "horreum.privacy")
     Optional<String> privacyStatement;
 
-    @ConfigProperty(name = "horreum.application.commit")
-    String appCommit;
+    @ConfigProperty(name = "horreum.application.commit", defaultValue = "<unknown>")
```
stalep, line 42: thank you!! :)

stalep (approved):

stalep:
I think this looks very good!
I actually started something similar after our chat about memory usage, but you dealt with almost all of it. I'll create a pr with some changes to Fingerprint as well which I think we can do.
squash and merge :)

barreiro (approved):
these look like a nice improvement. subtle changes that avoid loading state that is not really needed.

I approve !!

stalep:
> > TBH I don't get what's the purpose of it.. especially because when a dataset event is processed the dataset itself is already present so I don't get why it needs to be executed here but maybe I am missing something
> 
> I am actually thinking of removing the `validateDataset(newEvent.datasetId); ` from the `processDatasetEvents` method as the validation should have been run already, either during the upload or re-transformation, thus not needed during the dataset processing for the label values calculation/change detection. Am I missing something?

I've been looking a bit now and `validateDataset` is called by `RunServiceImpl.translate` so I do agree that this seems like it could be dropped.

Issue #2525, opened 2025-10-01
Listing datasets by Test fails when providing both `page` and `limit` with `page=0`


lampajr:
**Describe the bug**

Potential bug in the Horreum dataset API: `/api/dataset/list/{testId}` returns 500 when I use `limit` and `page` query params together. 

Works fine with either one alone, but combine them and it blows up:
`/api/dataset/list/262?limit=5` ✔️ works
`/api/dataset/list/262?page=0` ✔️  works
`/api/dataset/list/262?limit=5&page=0` ❌  500 error

Both params are in the OpenAPI spec as valid, so looks like a server-side issue.

From the server side I see the following stacktrace which leads to some malformed query:
```
(SELECT id FROM view WHERE test_id = ds.testid AND name = 'Default') WHERE testid = ? ORDER BY start DESC NULLS LAST limit 5 offset -5]
	at org.hibernate.exception.internal.SQLStateConversionDelegate.convert(SQLStateConversionDelegate.java:84)
	at org.hibernate.exception.internal.StandardSQLExceptionConverter.convert(StandardSQLExceptionConverter.java:34)
	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:115)
	at org.hibernate.sql.results.jdbc.internal.DeferredResultSetAccess.executeQuery(DeferredResultSetAccess.java:304)
	at org.hibernate.sql.results.jdbc.internal.DeferredResultSetAccess.getResultSet(DeferredResultSetAccess.java:200)
	at org.hibernate.sql.results.jdbc.internal.AbstractResultSetAccess.getResultSetMetaData(AbstractResultSetAccess.java:53)
	at org.hibernate.sql.results.jdbc.internal.AbstractResultSetAccess.getColumnCount(AbstractResultSetAccess.java:67)
	at org.hibernate.query.results.internal.ResultSetMappingImpl.resolve(ResultSetMappingImpl.java:188)
	at org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl.resolveJdbcValues(JdbcSelectExecutorStandardImpl.java:358)
	at org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl.resolveJdbcValuesSource(JdbcSelectExecutorStandardImpl.java:332)
[...]

PR #2526, opened 2025-10-01
Ignore page if < 1 and add description in the openapi

lampajr:
## Fixes Issue

Fixes https://github.com/Hyperfoil/Horreum/issues/2525

## Changes proposed

Passing `page <= 0` will be as not setting it at all (applied same logic used in other place for consistency)
 
- [ ] Improved the openapi doc by explicitly noting the `page` starts from `1`
- [ ] Ignore the `page` when `< 1` 

### DIFF
--- a/docs/site/content/en/openapi/openapi.yaml
+++ b/docs/site/content/en/openapi/openapi.yaml
@@ -3174,7 +3174,8 @@ paths:
         schema:
           type: integer
           format: int32
-      - description: filter by page number of a paginated list of Schemas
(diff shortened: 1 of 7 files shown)


stalep:
Looks very good!

stalep:
> Additionally we could add some validation in the endpoints using `page` and return a failure when trying to set it with a non-positive number.
> 
> I believe this could be done in a separate PR as it will require some more changes all over the codebase as many endpoints are implementing paging

Yeah, I think it should be a different pr.

PR #2527, opened 2025-10-01
Adapt custom JsonBinaryType to hibernate 7

lampajr:
## Fixes Issue

Fixes deprecation warning reported since Hibernate 7:
```
Overrides method that is deprecated and marked for removal in 'org.hibernate.usertype.UserType' 
```

## Changes proposed

- [x] As suggested in Hibernate java doc, override the other methods
- [x] Remove some overridden methods as they were matching the parent one

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/hibernate/JsonBinaryType.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/hibernate/JsonBinaryType.java
@@ -5,12 +5,11 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
-import java.util.Objects;
 
 import org.hibernate.HibernateException;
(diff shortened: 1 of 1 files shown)


stalep:
> > We are using `private final ObjectMapper mapper = new ObjectMapper();`, not sure whether having the mapper as `static` can be beneficial or not 🤔
> 
> Ok I don't think that it would make much difference

Hm, I think static could be useful. It should not be initiated more than once afaik...

stalep:
> > Hm, I think static could be useful. It should not be initiated more than once afaik...
> 
> yeah that's true, but as far as I can see in all places we are referencing a static instance
> 
> https://github.com/Hyperfoil/Horreum/blob/daebf6bf328e585f7c1b50a38861e23af6d67a6c/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/hibernate/JsonBinaryType.java#L23
> 
> , therefore I am expecting Hibernate will use the same static type. But anyway, I marked the mapper as static so we should be safer

Ah, ofc... Then it's probably fine yes...

PR #2532, opened 2025-10-02
Improve createFingerprint logic

lampajr:
Following up what I did in https://github.com/Hyperfoil/Horreum/pull/2522:

The previous logic, when it came to create a fingerprint, it was retrieving all label values for a dataset and applying further filtering based on the test.fingerprintLabels. This was creating a huge amount of allocation that was completely unnecessary, especially if the loaded label value was not a filtering label.

This was causing huge allocation rate and quite long time to even start the recalculation process for tests having many datasets with many (and big) label values. The process was getting stuck at https://github.com/Hyperfoil/Horreum/blob/e800b78afa9a328e15a9badea8689a93e9cfe192/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java#L1010

## Changes proposed

- [x] Retrieve only those label values that are in the fingerprint labels set
- [x] Batch the deletion to avoid many deletes one by one
- [x] Pass `fingerprintLabels` as parameter so that we can compute the list only once and not every execution 
- [x] Keep creating fingerprint even when the test fingerprint labels is empty

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
@@ -8,7 +8,6 @@ import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import java.util.stream.StreamSupport;
 
 import jakarta.annotation.security.PermitAll;
 import jakarta.enterprise.context.ApplicationScoped;
@@ -55,6 +54,7 @@ import io.hyperfoil.tools.horreum.hibernate.JsonBinaryType;
 import io.hyperfoil.tools.horreum.mapper.DatasetMapper;
 import io.hyperfoil.tools.horreum.server.WithRoles;
 import io.quarkus.logging.Log;
+import io.quarkus.panache.common.Parameters;
 import io.quarkus.runtime.Startup;
 import io.quarkus.security.identity.SecurityIdentity;
 
@@ -190,14 +190,6 @@ public class DatasetServiceImpl implements DatasetService {
              SELECT jsonb_array_elements_text(fingerprint_labels)
              FROM test WHERE id = :testId AND jsonb_typeof(fingerprint_labels) = 'array'
          );
-         """;
-
-    private static final String DELETE_FINGERPRINTS_BY_TEST = """
-         DELETE FROM fingerprint
-         WHERE dataset_id IN (
-            SELECT id FROM dataset
-            WHERE testid = :testId
-         )
          """;
     //@formatter:on
 
@@ -537,22 +529,6 @@ public class DatasetServiceImpl implements DatasetService {
         DataPointDAO.delete("dataset.id", datasetId);
     }
 
-    private List<String> getTestFingerprintLabelsAsList(int testId) {
-        List<String> fingerprintLabels = new ArrayList<>();
-        try {
-            JsonNode fingerprintLabelsJson = em
-                    .createQuery("SELECT t.fingerprintLabels from test t WHERE t.id = ?1", JsonNode.class)
-                    .setParameter(1, testId).getSingleResult();
-            if (fingerprintLabelsJson != null && fingerprintLabelsJson.isArray()) {
-                fingerprintLabels = StreamSupport.stream(fingerprintLabelsJson.spliterator(), false).map(JsonNode::asText)
-                        .toList();
-            }
-        } catch (NoResultException noResultException) {
-            Log.warnf("Could not find fingerprint for test: %d", testId);
-        }
-        return fingerprintLabels;
-    }
-
     @Transactional
     public void calcDatasetViews(int datasetId) {
         // TODO(user) move calc_dataset_view into Horreum business logic see https://github.com/hibernate/hibernate-orm/pull/7457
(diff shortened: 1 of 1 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
```diff
@@ -619,44 +636,40 @@ void createLabelValue(int datasetId, int labelId, JsonNode value) {
...
+        }
 
         FingerprintDAO fp = new FingerprintDAO();
         fp.datasetId = datasetId;
-        fp.dataset = DatasetDAO.findById(datasetId);
         fp.fingerprint = fpNode;
         fp.populateHash(); // this call is not really necessary
-        if (fp.datasetId > 0 && fp.dataset != null)
+        if (fp.datasetId > 0) {
             fp.persist();
```
franz1981, line 662: IN theory you could batch the created `FingerprintDao`s and let hibernate perform a batch insert too
(but I strongly suggest to decide for a specific batch size and probably you need to configure something on hibernate too?).

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
```diff
@@ -619,44 +636,40 @@ void createLabelValue(int datasetId, int labelId, JsonNode value) {
...
-        List<String[]> labelPairs = new ArrayList<>(labelValues.size());
-        for (var lv : labelValues)
-            labelPairs.add(new String[] { LabelDAO.<LabelDAO> findById(lv.labelId).name, lv.value.asText() });
-
-        for (int i = 0; i < json.size(); i++)
-            for (var name : labelPairs) {
-                if (json.get(i).asText().equals(name[0]))
-                    fpNode.put(name[0], name[1]);
-            }
+        List<Object[]> filteredLabelValues = session.createNativeQuery(LABEL_VALUES_BY_LABEL_NAMES, Object[].class)
```
franz1981, line 645: Why not retrieving all the filtered label values for all the dataset Ids you are interested in performing a single query to the DBMS?

franz1981: Clearly ChatGPT is not smart here, but this is what it said (wrongly very likely):
https://chatgpt.com/s/t_68df4b9f3f788191815157dd860b373c

lampajr: >Why not retrieving all the filtered label values for all the dataset Ids you are interested in performing a single query to the DBMS?

That's something I was considering as well (I am proceeding step by step haha), in this case I will need to implement batching as well because the number of datasets could be very large

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
lampajr: +1 that's very similar to the deletion, I believe we could batch it with proper sizing (technically something we can do on our side w/o additional configuration)

stalep:
I had a quick look at one of the failing tests; AlertingServiceTest.testChangeDetection
It seemes it's not creating a fingerprint and then the query on AlertingServiceImpl:633 returns null which says there are no more datapoints (if I understood it correctly).

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
```diff
@@ -619,44 +636,40 @@ void createLabelValue(int datasetId, int labelId, JsonNode value) {
...
     }
 
     @Transactional
     @WithRoles(extras = Roles.HORREUM_SYSTEM)
     void updateFingerprints(int testId) {
+        List<String> fingerprintLabels = getTestFingerprintLabelsAsList(testId);
         for (var dataset : DatasetDAO.<DatasetDAO> find("testid", testId).list()) {
             FingerprintDAO.deleteById(dataset.id);
-            createFingerprint(dataset.id, testId);
+            createFingerprint(dataset.id, fingerprintLabels);
```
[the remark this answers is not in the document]
lampajr: As found below, I need to slightly revert this as I need to create fingerprint even if the list is empty

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
lampajr: I did some trials and it's not that straightforward, so I'd like to postpone this improvement in other rounds

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
```diff
@@ -515,13 +518,27 @@ void calculateLabelValues(int testId, int datasetId, Integer[] queryLabelIds) {
...
         // label values have been recomputed, invalidate existing datapoints
         // cleanup datapoints for the current dataset
         DataPointDAO.delete("dataset.id", datasetId);
     }
 
+    private List<String> getTestFingerprintLabelsAsList(int testId) {
+        List<String> fingerprintLabels = new ArrayList<>();
+        try {
+            JsonNode fingerprintLabelsJson = em
+                    .createQuery("SELECT t.fingerprintLabels from test t WHERE t.id = ?1", JsonNode.class)
```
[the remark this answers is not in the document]
lampajr: I was able to implement that with commit https://github.com/Hyperfoil/Horreum/pull/2532/commits/e4cd1bc41d192ea25d9e7e205f0817174797db2b, this way no need to have a separate query to compute such list but it is performed as inner select in the query.

lampajr:
Hi @stalep @barreiro, I believe this other round of improvements is ready to be reviewed. 

This mainly focused on improving the createFingerprint method, especially in the memory allocation. As you can see from the following charts, I was able to reduce the memory allocation quite a lot during the "synchronous" datapoint recalculation when dealing with big Horreum tests:

**Before (current master)**
image

**This PR**
image

Moreover the time needed to re-compute all fingerprints for the same big test went from "more than 30 mins" to "less than 2 mins"

barreiro (approved):
I'm leaving a few suggestions. it's mostly form over functionality.

lampajr:
As agreed offline, going to merge this one!

PR #2540, opened 2025-10-07
Fix label selection when dealing with numbers

lampajr:
## Fixes Issue

Fixing issue when the fingerprint values are numbers, in this situation the strict equality won't work as number <> string.

## Changes proposed

- [ ] Use not strict equality (` } else if (ls[key] != value) {` instead of ` } else if (ls[key] !== value) {`)
- [ ] Remove some unused code


lampajr:
Let's put this on hold since I believe I might have changed how fingerprints are created, now keeping the original value instead of the text representation which might have caused the problem

lampajr:
> Let's put this on hold since I believe I might have changed how fingerprints are created, now keeping the original value instead of the text representation which might have caused the problem

It should be fixed now, the problem (as I was suspecting) was that in current prod version we are creating the fingerprints with values in text form (see https://github.com/Hyperfoil/Horreum/blob/954c2b22eab0bdad9a5c3fca649e551089b08f6d/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java#L576C90-L576C107) but with the recent changes that was inadvertently changed.

@barreiro now the PR is ready to be reviewed again, thanks for the patience 🙏🏻

lampajr:
> Interesting, now getting one test failing:
> 
> ```
> AlertingServiceTest.testFingerprintLabelsChange:255 expected: <{"foo":"aaa"}> but was: <{"foo":"\"aaa\""}>
> ```
> 
> Looking into it 🤦🏻

I've updated the test so that we have both text and numeric labels where both are converted to text properly

PR #2555, opened 2025-10-17
Return filtering label values as map

lampajr:
## Fixes Issue

Fixes the issue that users might experience when they have multiple filtering labels values and many datasets.

Currently the `filteringLabelValues` endpoint returns for each dataset an object with the values of all filtering labels (defined in the schema on that dataset) - this could end up in loading a lot of data and making the browser hanging!

## Changes proposed

- [x] Add new endpoint to avoid introducing breaking change (`/test/{0}/filteringLabels`) 
- [x] Return a map where the key is the name of the label and the value is the list of all possible values for that label (across all datasets)
- [x] Adapt the UI to manage this result structure
- [x] Mark as deprecated the old endpoint

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
@@ -653,20 +653,13 @@ public class TestServiceImpl implements TestService {
         // This check is good, keep it
         TestDAO.findByIdOptional(testId).orElseThrow(() -> ServiceException.serverError("Cannot find test " + testId));
 
-        List<Object[]> filters = em.createNativeQuery(LABEL_VALUES_SUMMARY_QUERY_NEW)
+        return em.unwrap(Session.class).createNativeQuery(LABEL_VALUES_SUMMARY_QUERY_NEW, Object[].class)
                 .setParameter("testId", testId)
-                .unwrap(NativeQuery.class)
                 .addScalar("name", StandardBasicTypes.STRING)
                 .addScalar("val", JsonBinaryType.INSTANCE)
-                .getResultList(); // No need to unwrap
-
-        if (filters == null) {
-            return new HashMap<>();
-        }
-
-        return filters.stream().collect(Collectors.toMap(
-                row -> (String) row[0], // name of the label as key
-                row -> (JsonNode) row[1]));
+                .stream().collect(Collectors.toMap(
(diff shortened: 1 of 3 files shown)


horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/TestServiceImpl.java
```diff
@@ -637,6 +649,26 @@ public List<ObjectNode> filteringLabelValues(
...
+                .addScalar("val", JsonBinaryType.INSTANCE)
+                .getResultList(); // No need to unwrap
+
+        if (filters == null) {
+            return new HashMap<>();
+        }
+
+        return filters.stream().collect(Collectors.toMap(
+                row -> (String) row[0], // name of the label as key
+                row -> (JsonNode) row[1]));
```
barreiro:
```suggestion
        return em.unwrap(Session.class).createNativeQuery(LABEL_VALUES_SUMMARY_QUERY_NEW, Object[].class)
                .setParameter("testId", testId)
                .addScalar("name", StandardBasicTypes.STRING)
                .addScalar("val", JsonBinaryType.INSTANCE)
                .stream().collect(Collectors.toMap(
                        row -> (String) row[0], // name of the label as key
                        row -> (JsonNode) row[1]));
```

horreum-web/src/domain/runs/TestDatasets.tsx
```diff
@@ -182,12 +183,11 @@ export default function TestDatasets() {
...
             }}
         >
-            <ToolbarContent>
+            <ToolbarContent alignItems={"center"}>
                 <LabelFilter
                     selection={filter}
                     onSelect={setFilter}
-                    source={labelsSource}
-                    emptyPlaceholder={<span>No filters available</span>}
+                    source={newLabelsSource}
```
barreiro:
```suggestion
                    source={labelsSource}
```

horreum-web/src/components/LabelFilter/LabelFilter.tsx
```diff
@@ -53,105 +32,57 @@ type LabelsSelectProps = {
...
-                        if (deepEquals(value, values[i])) {
-                            return false
-                        }
-                    }
-                    return true
-                })
-                .map(value => convertLabelValue(value))
+        const items = Object.entries(availableLabels).map(label => {
+            const opts = label[1]
+                .map((value: any) => convertLabelValue(value))
```
barreiro:
```suggestion
                .map(convertLabelValue)
```

PR #2559, opened 2025-10-20
this usually triggers a deadlock...

stalep:
Not always, but usually this will trigger a deadlock locally.

```
2025-10-20 19:28:58,606 INFO  [io.hyp.too.hor.svc.NotificationServiceImpl] (executor-thread-1) Received new changes in test 10 (quarkus-spring-boot-comparison), dataset 12/2 (fingerprint: {"Runtime":"spring3","buildType":"native"})
2025-10-20 19:28:58,713 WARN  [org.hib.orm.jdb.error] (horreum.dataset.pool-10) HHH000247: ErrorCode: 0, SQLState: 40P01
2025-10-20 19:28:58,713 WARN  [org.hib.orm.jdb.error] (horreum.dataset.pool-10) ERROR: deadlock detected
  Detail: Process 74 waits for ShareLock on transaction 1881; blocked by process 68.
Process 68 waits for AccessExclusiveLock on tuple (6,6) of relation 16621 of database 16384; blocked by process 65.
Process 65 waits for ShareLock on transaction 1891; blocked by process 74.
  Hint: See server log for query details.
  Where: while deleting tuple (3,5) in relation "change"
2025-10-20 19:28:58,718 ERROR [io.sma.rea.mes.provider] (horreum.dataset.pool-10) SRMSG00200: The method io.hyperfoil.tools.horreum.svc.ServiceMediator#processDatasetEvents has thrown an exception: jakarta.persistence.PessimisticLockException: JDBC exception executing SQL [ERROR: deadlock detected
  Detail: Process 74 waits for ShareLock on transaction 1881; blocked by process 68.
Process 68 waits for AccessExclusiveLock on tuple (6,6) of relation 16621 of database 16384; blocked by process 65.
Process 65 waits for ShareLock on transaction 1891; blocked by process 74.
  Hint: See server log for query details.
  Where: while deleting tuple (3,5) in relation "change"] [DELETE FROM change cc
WHERE NOT cc.confirmed
  AND cc.variable_id = ?
  AND (cc.timestamp > ? OR (cc.timestamp = ? AND ?))
  AND EXISTS (
      SELECT 1 FROM fingerprint fp
      WHERE fp.dataset_id = cc.dataset_id
        AND fp.fp_hash = ?
  );
]
	at org.hibernate.internal.ExceptionConverterImpl.wrapLockException(ExceptionConverterImpl.java:227)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:85)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:155)
	at org.hibernate.query.spi.AbstractQuery.executeUpdate(AbstractQuery.java:647)
	at io.hyperfoil.tools.horreum.svc.AlertingServiceImpl.runChangeDetection(AlertingServiceImpl.java:658)
```

### DIFF
--- a/horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TestServiceTest.java
+++ b/horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TestServiceTest.java
@@ -642,39 +642,44 @@ class TestServiceTest extends BaseServiceTest {
         assertEquals(11, labels2.size());
 
         BlockingQueue<Test> newDataPoints = serviceMediator.getEventQueue(AsyncEventChannels.DATAPOINT_NEW, testId);
+        BlockingQueue<Test> updates = serviceMediator.getEventQueue(AsyncEventChannels.DATASET_UPDATED_LABELS, testId);
         List<Integer> runIds = new ArrayList<>();
         for (int i = 1; i < 5; i++) {
+            for (int j = 1; j < 11; j++) {
+                BlockingQueue<Test> events = serviceMediator.getEventQueue(AsyncEventChannels.RUN_NEW, testId);
+                Run run = new Run();
+                run.testid = testId;
+                run.data = mapper.readTree(p.resolve("quarkus_sb_run" + i + ".json").toFile());
+                run.start = Instant.parse(run.data.get("timing").get("start").asText()).plusSeconds(j);
+                run.stop = Instant.parse(run.data.get("timing").get("stop").asText()).plusSeconds(j);
+                run.owner = "foo-team";
 
-            BlockingQueue<Test> events = serviceMediator.getEventQueue(AsyncEventChannels.RUN_NEW, testId);
-            Run run = new Run();
-            run.testid = testId;
-            run.data = mapper.readTree(p.resolve("quarkus_sb_run" + i + ".json").toFile());
-            run.start = Instant.parse(run.data.get("timing").get("start").asText());
-            run.stop = Instant.parse(run.data.get("timing").get("stop").asText());
-            run.owner = "foo-team";
+                Response response = jsonRequest()
+                        .auth()
+                        .oauth2(getUploaderToken())
+                        .body(run)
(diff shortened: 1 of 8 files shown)


horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/TestServiceTest.java
```diff
@@ -613,6 +614,7 @@ private TestService.TestQueryResult listTests(String token, String roles, Intege
     }
 
     @org.junit.jupiter.api.Test
+    @Ignore
```
lampajr: If you want to ignore a test I think you should use `@Disabled` instead of `@Ignore`. But I was wondering why this started to fail, is it because of some internal changes due to the removal of the `executeBlocking` ?

stalep: I need to verify, but the recalculation seemed to take longer and the sleep I had previously was not enough to match the expected labelValues. I added a poller now which hopefully should work better.

lampajr (changes requested):
Hi @stalep,

I did another quick round and from a correctness standpoint it seems working as expected but there are a couple of concerns:

* If you trigger the recalculation in the Change Detection tab you will see the bar will not progress (I believe because the progress is not properly tracked in the backend)
image


* The process of recalculating datasets (consuming messages) seems **much much slower**. This is my biggest concern, I am trying some tests locally with a prod db to compare before/after to have some broad numbers..

stalep:
Yeah, I did notice some slowdown as well, thanks for confirming. Before I had the insight I had now, I wanted to make sure we did it "correct" as I thought the deadlock could be related to this. However, I now know it's related to the changer calculation and not a problem with doing the datapoint calculations in parallel.
I would however like if we can code this without calling vertx.blabla directly (sorry, in a rush we can discuss more later)

PR #2568, opened 2025-10-29
Fix how "all matches" extractor are handled

lampajr:
## Fixes Issue

The expectation is that "previewing" a label execution and its actual label value computation return the same result, which is not true a the moment when one of the extractor is marked as "all matches".

## Changes proposed

- [x] Removed apparently wrong `flatten((ArrayNode) row[3])`
- [x] Added some more tests

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
@@ -520,7 +520,7 @@ public class DatasetServiceImpl implements DatasetService {
         FingerprintDAO.deleteById(datasetId);
         Util.evaluateWithCombinationFunction(toCompute,
                 (row) -> (String) row[2],
-                (row) -> (row[3] instanceof ArrayNode ? flatten((ArrayNode) row[3]) : (JsonNode) row[3]),
+                (row) -> (JsonNode) row[3],
                 (row, result) -> createLabelValue(datasetId, (int) row[0], Util.convertToJson(result)),
                 (row) -> createLabelValue(datasetId, (int) row[0], (JsonNode) row[3]),
                 (row, e, jsCode) -> logMessage(datasetId, PersistentLogDAO.ERROR,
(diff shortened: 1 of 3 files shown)


lampajr:
Added some more tests covering first and all matches scenarios , especially when dealing with one single extractor.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
```diff
@@ -520,7 +520,7 @@ void calculateLabelValues(int testId, int datasetId, Integer[] queryLabelIds) {
         FingerprintDAO.deleteById(datasetId);
         Util.evaluateWithCombinationFunction(toCompute,
                 (row) -> (String) row[2],
-                (row) -> (row[3] instanceof ArrayNode ? flatten((ArrayNode) row[3]) : (JsonNode) row[3]),
```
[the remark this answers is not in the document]
stalep: Yes, idk if we have any data that would be affected if we remove it?

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/DatasetServiceTest.java
```diff
@@ -546,4 +547,202 @@ public void testDatasetListByRun() {
...
+        barExtractor.name = "bar";
+        barExtractor.jsonpath = "$.bar";
+
+        // pass the array of arrays to a js function
+        Label label = createLabelDto(fooSchema, "labelFoo", "val => val.foo", null, fooExtractor, barExtractor);
+        Response addLabelResponse = jsonRequest().body(List.of(label)).post("/api/schema/" + fooSchema.id + "/labels");
+        addLabelResponse.then().statusCode(201);
+
+        List<Integer> ids = uploadRun(
+                "{ \"bar\": \"test1\", \"foo\": [{\"inner\":[{\"id\": 1}, {\"id\": 2}]}, {\"inner\":[{\"id\": 3}]}, {\"inner\":[{\"id\": 4}, {\"id\": 5}]}]}",
```
willr3: can we use `"""` to make this easier to read?

horreum-backend/src/test/java/io/hyperfoil/tools/horreum/svc/DatasetServiceTest.java
```diff
@@ -546,4 +547,202 @@ public void testDatasetListByRun() {
...
+        fooExtractor.jsonpath = "$.foo[*].inner";
+        fooExtractor.isArray = false;
+
+        // pass the array of arrays to a js function
+        Label label = createLabelDto(fooSchema, "labelFoo", "val => val", null, fooExtractor);
+        Response addLabelResponse = jsonRequest().body(List.of(label)).post("/api/schema/" + fooSchema.id + "/labels");
+        addLabelResponse.then().statusCode(201);
+
+        List<Integer> ids = uploadRun(
+                "{ \"bar\": \"test1\", \"foo\": [{\"inner\":[{\"id\": 1}, {\"id\": 2}]}, {\"inner\":[{\"id\": 3}]}, {\"inner\":[{\"id\": 4}, {\"id\": 5}]}]}",
```
willr3: same comment about readability

lampajr:
> Only made minor comments about readability and the background on the `async` test I added

Applied your suggestions ✅

PR #2572, opened 2025-10-30
Refactor oidc auth management

lampajr:
## Fixes Issue

Fixes current issue with Keycloak auth session management in the UI

## Changes proposed

- [x] Huge refactoring of the OIDC management leveraging https://github.com/authts/react-oidc-context
- [x] Complete removal of `react-redux`
- [x] Remove dependency on `keycloak-js`

### DIFF
--- a/horreum-web/src/App.tsx
+++ b/horreum-web/src/App.tsx
@@ -42,12 +42,13 @@ import About from "./About"
 import AppContextProvider from "./context/AppContext";
 import TableReportConfigPage from "./domain/reports/TableReportConfigPage";
 import TableReportPage from "./domain/reports/TableReportPage";
-import {createUserManager, onSigninCallback} from "./auth/oidc";
+import {createUserManager} from "./auth/oidc";
 import {AuthProvider} from "react-oidc-context";
 import {KeycloakConfig} from "./generated";
 import {configApi} from "./api";
 import AuthBridgeContextProvider, {AuthBridgeContext} from "./context/AuthBridgeContext";
 import {AuthContextType} from "./context/@types/authContextTypes";
+import {User} from "oidc-client-ts";
 import CallbackSSO from "./auth/CallbackSSO";
 
 const router = createBrowserRouter(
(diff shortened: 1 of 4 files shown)


horreum-web/package.json
```diff
@@ -27,11 +26,12 @@
   "luxon": "3.3.0",
   "moment": "2.30.1",
   "monaco-editor": "0.54.0",
+  "oidc-client-ts": "^3.3.0",
```
barreiro: we can remove the `keycloak-js` dependency :)

horreum-web/src/auth/oidc.ts
```diff
@@ -0,0 +1,19 @@
+import {KeycloakConfig} from "../generated";
+import {UserManager, WebStorageStateStore} from "oidc-client-ts";
+
+export const createUserManager = (config: KeycloakConfig): UserManager => {
+    return new UserManager({
+        authority: config.url + "/realms/horreum",
+        client_id: config.clientId ?? "",
+        redirect_uri: `${window.location.origin}/silent-check-sso`,
```
barreiro: this is not correct. this value is for the `silent_redirect_uri` property.

`redirect_uri` it should be 
```
redirect_uri: `${window.location.origin}/${window.location.pathname}`
```
(assuming the OIDC provider accepts all redirects, which I believe it does). if not the alternative is to track the pathname in local state

the idea here is when the user press login from some page (say a test or a dataset page) it gets redirected to that same page after login.

horreum-web/package.json
lampajr: +1 forgot to remove it

horreum-web/src/App.tsx
```diff
@@ -84,56 +90,35 @@ const router = createBrowserRouter(
 );
 
 export default function App() {
-    initKeycloak(store.getState())
+    const [horreumOidcConfig, setHorreumOidcConfig] = useState<KeycloakConfig | undefined>()
+
+    useEffect(() => {
+        configApi.keycloak().then(setHorreumOidcConfig)
```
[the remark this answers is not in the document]
lampajr: Interesting idea, I can try to take a look at that and if it does not add too much effort I agree on adding that as part of this change.. otherwise I am also fine to defer that change to another pr

horreum-web/src/auth/oidc.ts
lampajr: I am still banging my head trying to make it work the proper way but using 
```
redirect_uri: `${window.location.origin}/${window.location.pathname}`
// or
redirect_uri: `${window.location.origin}${window.location.pathname}`
```

does not work, the authentication itself works and it redirects back but.. for some reasons the context does not recognize the authentication to be succeeded

horreum-web/src/App.tsx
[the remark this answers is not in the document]
lampajr: I would defer this additional improvement to a separate PR to avoid adding too much changes here

lampajr:
> @barreiro you mentioned you had some comments, please add them asap or we'll merge...

Rebased and resolved the package.json conflicts, is there anything I should address here? or do you think it can be merged in master? and not sure whether we want to backport it

lampajr:
As agreed offline with @stalep, merging this one as it is right now. Further improvements can be added later on 🚀

PR #2578, opened 2025-11-07
Changes to recalculation

barreiro:
## Fixes Issue

Improves how the change detection is performed.

This PR incorporates #2559  

## Changes proposed

A new event and queue is used to decouple change detection from datapoint creation. 
Change detection now is done in one go, instead of the previous algorithm.

- [ x] My code follows the code style of this project.
- [x ] All new and existing tests passed.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/AlertingServiceImpl.java
@@ -12,7 +12,6 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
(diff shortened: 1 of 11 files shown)


stalep:
Is this ready to be reviewed @barreiro ?

barreiro:
rebased and fixed the failing test... it's ready now!

lampajr (changes requested):
I just started reviewing it and I found this:

- When you click "save" on "Change detection" tab, you can recalculate all datapoints but since now the computation is more synchronous the progress bar does not reflect the real status it goes from 0 to 100 as soon as completed
image

I believe this should be properly adapted to the new logic.

I will keep reviewing it..

lampajr:
I did some other tests trying to recalculate datapoints/changes and the first results look, so I believe that once the issue I shared is sorted out we could try to merge it and you can see in staging (with ~prod like data)

PR #2620, opened 2026-07-15
Feat: add ChangeToSlackWebhook formatter for HttpAction

vishnandaman:
## Problem

  The `slack-channel-message` action requires a bot token with `chat:write` scope.
  In enterprise Slack workspaces, adding this scope requires an IT approval process
  that can take weeks.

  Users who already have a Slack app with `incoming-webhook` (auto-approved) cannot
  use the `http` action as a fallback — it sends raw Horreum JSON that Slack's
  Incoming Webhook API rejects.

  ## Solution

  Add an optional `formatter` field to the `http` action. When set, the payload is
  formatted and wrapped in `{"text": "..."}` before posting — exactly what Slack
  webhooks expect. When absent, behavior is unchanged (raw JSON).

  ```json
  {
    "event": "change/new",
    "type": "http",
    "config": {
      "url": "https://hooks.slack.com/services/...",
      "formatter": "changeToMarkdown"
    }
  }
```

 ## Changes

  - HttpActionConfig.java — optional formatter field
  - HttpAction.java — format and wrap payload when formatter is configured
  - ActionComponentForm.tsx — formatter dropdown for http action in UI

## Backwards compatibility

  Fully backwards compatible — existing http actions without formatter are unaffected.

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/action/ChangeToSlackWebhook.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/action/ChangeToSlackWebhook.java
@@ -1,0 +1,56 @@
+package io.hyperfoil.tools.horreum.action;
+
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import io.hyperfoil.tools.horreum.api.alerting.Change;
+import io.hyperfoil.tools.horreum.entity.data.DatasetDAO;
+import io.hyperfoil.tools.horreum.svc.Util;
+import io.quarkus.qute.Location;
+import io.quarkus.qute.Template;
+
+@ApplicationScoped
+public class ChangeToSlackWebhook implements BodyFormatter {
+    @Location("new_issue_from_change")
+    Template template;
+
+    @ConfigProperty(name = "horreum.url")
+    String publicUrl;
+
+    @Override
+    public String name() {
+        return "changeToSlackWebhook";
+    }
+
+    @Override
+    public String format(JsonNode config, Object payload) {
+        if (!(payload instanceof Change.Event)) {
+            throw new IllegalArgumentException("This formatter accepts only Change.Event!");
+        }
+        Change.Event event = (Change.Event) payload;
+        Change change = event.change;
+        String fingerprint = DatasetDAO.getEntityManager().getReference(DatasetDAO.class, change.dataset.id).getFingerprint();
+        String text = template
+                .data("testName", event.testName)
+                .data("testNameEncoded", URLEncoder.encode(event.testName, StandardCharsets.UTF_8))
+                .data("fingerprint", URLEncoder.encode(fingerprint, StandardCharsets.UTF_8))
+                .data("publicUrl", publicUrl)
+                .data("testId", String.valueOf(change.variable.testId))
+                .data("variable", change.variable.name)
+                .data("group", change.variable.group)
+                .data("runId", event.change.dataset.runId)
+                .data("datasetOrdinal", event.change.dataset.ordinal)
+                .data("description", change.description)
+                .render();
+        ObjectNode body = Util.OBJECT_MAPPER.createObjectNode();
+        body.put("text", text);
+        return body.toString();
+    }
+}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: horreum-web/src/domain/actions/ActionComponentForm.tsx [hunk 1]
```diff
@@ -118,15 +118,36 @@ export default function ActionComponentForm(props: ActionComponentFormProps) {
...
+                                (props.action.event === CHANGE_NEW
+                                    ? [{value: "changeToMarkdown", content: "Change to Markdown"}]
+                                    : props.action.event === EXPERIMENT_RESULT_NEW
+                                        ? [{value: "experimentsResultToMarkdown", content: "Experiment result to Markdown"}]
+                                        : props.action.event === TEST_NEW
+                                            ? [{value: "testToSlack", content: "Test to Slack Markdown"}]
+                                            : []
+                                ).map(o => ({...o, selected: o.value === (props.action.config as Http).formatter}))
+                            }
+                            placeholderText="None (send raw JSON)"
```
stalep: This seems to cause a build error as PatternFly6 only accepts `placeHolder` not `placeHolderText`.

stalep (commented):
The "main" concern I have with this is that it's hardcoding Slack-specific ("text": "..." } wrapping to a generic class...

horreum-web/src/domain/actions/ActionComponentForm.tsx
vishnandaman: Thank you @stalep for feedback!
Addressed both concerns :
- I have moved the Slack specific { "text" : "..."  } wrapping out of the HttpAction into new ChangeToSlackWebhook formatter.
- fixed the PatternFlay 6 issue for placeholder

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/action/ChangeToSlackWebhook.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.hyperfoil.tools.horreum.api.alerting.Change;
+import io.hyperfoil.tools.horreum.entity.data.DatasetDAO;
+import io.hyperfoil.tools.horreum.svc.Util;
+import io.quarkus.qute.Location;
+import io.quarkus.qute.Template;
+
+@ApplicationScoped
+public class ChangeToSlackWebhook implements BodyFormatter {
+    @Location("new_issue_from_change")
```
stalep: The `new_issue_from_change` Qute template, which uses standard Markdown link syntax:
`[{testName}]({publicUrl}/test/{testId})`
Slack doesn't support standard Markdown — it uses its own mrkdwn format:
`<{publicUrl}/test/{testId}|{testName}>`
A dedicated Slack template is needed (e.g., slack_from_change.md), following the same pattern as the existing TestToSlackMarkdown which uses slack_from_test.md. Without this, links will render as raw text in Slack messages.

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/action/ChangeToSlackWebhook.java
```diff
@@ -0,0 +1,56 @@
...
+    @ConfigProperty(name = "horreum.url")
+    String publicUrl;
+
+    @Override
+    public String name() {
+        return "changeToSlackWebhook";
+    }
+
+    @Override
+    public String format(JsonNode config, Object payload) {
```
stalep, line 31: All other BodyFormatter implementations return raw formatted text. The JSON envelope wrapping is handled by the action plugin, not the formatter. If this formatter were used via SlackChannelMessageAction it would double-wrap and produce garbage. The {"text": "..."} wrapping should happen in HttpAction.buildBody() instead, and the formatter should return plain text like all the others.

horreum-web/src/domain/actions/ActionComponentForm.tsx
```diff
@@ -118,15 +118,39 @@ export default function ActionComponentForm(props: ActionComponentFormProps) {
...
+                    <FormGroup label="Formatter" fieldId="formatter">
+                        <SimpleSelect
+                            initialOptions={
+                                (props.action.event === CHANGE_NEW
+                                    ? [
+                                        {value: "changeToMarkdown", content: "Change to Markdown"},
+                                        {value: "changeToSlackWebhook", content: "Change to Slack Webhook"}
+                                      ]
+                                    : props.action.event === EXPERIMENT_RESULT_NEW
+                                        ? [{value: "experimentsResultToMarkdown", content: "Experiment result to Markdown"}]
```
stalep: The frontend dropdown has "experimentsResultToMarkdown" (with an "s") but the backend ExperimentResultToMarkdown uses "experimentResultToMarkdown" (no "s"). This is a pre-existing bug, but since you're touching this code it would be good to fix it rather than propagate it.

stalep (commented):
Thanks for the update, the `HttpAction.buildBody()` is now generic which is a good improvement, and the PatternFly 6 placeholder fix is in. A few things still need to be addressed:

horreum-web/src/domain/actions/ActionComponentForm.tsx
vishnandaman: I have removed the 's' to maintain the consistency Thanks for pointing out this !

horreum-backend/src/main/java/io/hyperfoil/tools/horreum/action/ChangeToSlackWebhook.java
vishnandaman: Oh I see... I have moved the text wrapping then to buildBody of HttpAction only for slack webhooks

PR #2621, opened 2026-07-16
optimize dataset listing query

barreiro:
The `all_labels` CTE in `listByTest()` and `listByRun()` was aggregating ALL label values for every dataset into a JSON object, then filtering with `@>` containment. For large tests this materializes tens of millions of rows just to match against a filter with one or two keys.

The fix restricts the CTE to only aggregate labels whose names appear in the filter (`label.name IN (SELECT jsonb_object_keys(:jsonFilter))`), reducing intermediate rows by ~387x in the observed case.

## Test plan

- [ ] Verify `listByTest` with a label filter returns correct results
- [ ] Verify `listByTest` without a filter is unaffected
- [ ] Verify `listByRun` with and without filter

### DIFF
--- a/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
+++ b/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/svc/DatasetServiceImpl.java
@@ -178,8 +178,8 @@ public class DatasetServiceImpl implements DatasetService {
          SELECT dataset.id as dataset_id,
             COALESCE(jsonb_object_agg(label.name, lv.value) FILTER (WHERE label.name IS NOT NULL), '{}'::jsonb) AS values
          FROM dataset
-         LEFT JOIN label_values lv ON dataset.id = lv.dataset_id
-         LEFT JOIN label ON label.id = label_id
+         JOIN label_values lv ON dataset.id = lv.dataset_id
+         JOIN label ON label.id = lv.label_id
          """;
 
     private static final String GET_FINGERPRINT_FROM_LABEL_VALUES = """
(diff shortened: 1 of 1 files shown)


diegolovison:
### Minor Considerations to Double-Check

1. **Top-Level Keys Only:** PostgreSQL's `jsonb_object_keys()` only returns the *top-level* keys of the JSON object. This assumes that `filter` (parsed via `Util.parseFingerprint(filter)`) will always be a flat JSON object (e.g., `{"label1": "value", "label2": "value"}`). If you ever introduce nested JSON filters, this logic will miss the nested keys.

Looking at the specific filter in your log:
`?filter={".metadata.env.BUILD_ID": "Jenkins..."}`

Even though there are dots in the string, this is a **flat** JSON object with a single top-level key named `".metadata.env.BUILD_ID"`. Because you aren't passing a deeply nested JSON object (e.g., `{"metadata": {"env": {"BUILD_ID": "..."}}}`), `jsonb_object_keys()` will perfectly extract `".metadata.env.BUILD_ID"` and filter correctly. As long as your application continues to structure fingerprints as flat key-value pairs where the key is the JSONPath/label name, this optimization will work flawlessly.

2. **Indexing:** To get the absolute maximum benefit out of this query, make sure there is an index on `label(name)` in your database schema. However, even without an index, dropping rows before `jsonb_object_agg` is still a net win for memory and CPU overhead.
3. **Left Joins in `ALL_LABELS_SELECT`:** The base `ALL_LABELS_SELECT` uses `LEFT JOIN label_values` and `LEFT JOIN label`. By adding a `WHERE label.name IN (...)` condition to the CTE, you are implicitly converting that `LEFT JOIN` on `label` into an `INNER JOIN` (because `label.name` can no longer be null). Since you are specifically looking for datasets that *contain* these labels to match the filter anyway, this side effect is actually perfectly fine and logically correct, but it's worth noting. Change to `INNER JOIN`

barreiro:
1. yes, fingerprints are always flat JSON objects. the keys are the label names (like `".metadata.env.BUILD_ID"`) and `jsonb_object_keys()` extracts the value correctly.
2. there's already a unique index `label_name_schema_id_key` on `label(name, schema_id)` that can be used for the `label.name IN (...)` lookup.
3. good point. I updated `ALL_LABELS_SELECT` to use `INNER JOIN` since the filter branches require matching labels. this makes the intent explicit and gives the query planner better information.

diegolovison (approved):
LGTM

